PR #1, opened 2016-06-02
Setup dot files

dreispt:


oca_dependencies.txt
```diff
@@ -0,0 +1,15 @@
...
+#
+# To depend on the standard version of sale-workflow, use:
+# sale-workflow
+#
+# To explicitely give the URL of a fork, and still use the version specified in
+# .travis.yml, use:
+# sale-workflow https://github.com/OCA/sale-workflow
+#
+# To provide both the URL and a branch, use:
+# sale-workflow https://github.com/OCA/sale-workflow branchname
```
elicoidal, line 15: @dreispt 
I would add at least project and project-reporting

elicoidal, line 15: Other than that, :+1: 
Thanks for the maintenance!

elicoidal:
BTW, I have another issue: I dont have the same labels as usual. Do you know where it should be fixed?

oca_dependencies.txt
dreispt, line 15: I would prefer to have the PRs introducing dependencies doing that.

dreispt:
I don't have access to add Labels. I think there's a script to add them, but not familiar with the process.

dreispt:
(I'll forward port this to 9.0 once merged)

oca_dependencies.txt
elicoidal, line 15: good to me

dreispt:
Sorry for the rush, I'm merging so that PRs can take advantage of TravisCI checks.

PR #25, opened 2016-09-23
9.0 16334 migrate business requirement

sudhir-serpentcs:
Migrated business_requirement module from v8 to v9.
@dreispt  @pedrobaeza @jbeficent @moylop260 Please review.


business_requirement/README.rst
```diff
@@ -0,0 +1,271 @@
...
+
+Sub-business requirements
+-------------------------
+User can create sub business requirements for a given BR for complex cases.
+
+This is a simple parent/child relation (see roadmap).
+
+.. figure:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/222/8.0
```
elicoidal: 9.0

business_requirement/README.rst
```diff
@@ -0,0 +1,271 @@
...
+   * Master Project (Partner is the one related to the master project)
+   * Priority (low, Medium, High)
+   * Change request: Is it a change request? (currently only informational with n model or action)
+   * Tags: any relevant tag for the business.
+   * Owner and approver by default
+   
+   .. figure:: ../business_requirement/static/img/bus_req_tags2.png
+      :width: 600 px
+
+   .. figure::business_requirement/static/img/bus_req_tags2.png
```
elicoidal: For all pictures, keep only one remove business_requirement/

business_requirement/models/business.py
```diff
@@ -0,0 +1,275 @@
...
+class BusinessRequirement(models.Model):
+    _inherit = ['mail.thread', 'ir.needaction_mixin']
+    _name = "business.requirement"
+    _description = "Business Requirement"
+    _order = 'id desc'
+
+    @api.model
+    def _get_default_company(self):
+        company_id = self.env.user._get_company()
+        if not company_id:
```
elicoidal: I think this should never happen as company is mandatory.

elicoidal (changes requested):
some details in README . :+1:

elicoidal:
tested in runbot: :+1:

business_requirement/models/business.py
```diff
@@ -0,0 +1,275 @@
...
+    )
+    approved_id = fields.Many2one(
+        'res.users',
+        string='Approved by',
+        readonly=True
+    )
+    company_id = fields.Many2one(
+        'res.company', string='Company',
+        required=True, readonly=True, states={'draft': [('readonly', False)]},
+        default=_get_default_company,
```
moylop260: What about use the following way https://github.com/odoo/odoo/blob/b2804b1d2ae5fd7ec5c0eca79a8db77476badc84/addons/account/models/account.py#L192?

`company_id = fields.Many2one('res.company', string='Company', required=True, index=True, default=lambda self: self.env.user.company_id,
        help="Company related to this journal")`

business_requirement/models/business.py
```diff
@@ -0,0 +1,275 @@
...
+    )
+
+    @api.multi
+    @api.onchange('project_id')
+    def project_id_change(self):
+        if self.project_id and self.project_id.partner_id:
```
moylop260: What about use:

``` diff
- if self.project_id and self.project_id.partner_id:
-      self.partner_id = self.project_id.partner_id.id
+ self.partner_id = self.project_id.partner_id.id
```

Now the new api support `BrowseNull.BrowseNull` and IMHO this onchange should change in order to avoid use a old value of other onchange.

business_requirement/models/business.py
```diff
@@ -0,0 +1,275 @@
...
+            content_subtype='html', **kwargs
+        )
+        return res
+
+
+class BusinessRequirementCategory(models.Model):
+    _name = "business.requirement.category"
+    _description = "Categories"
+
+    name = fields.Char(string='Name', required=True)
```
moylop260: FYI https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#field
`If the technical name of the field (the variable name) is the same to the string of the label, don't put string parameter for new API fields, because it's automatically taken`

business_requirement/models/business.py
sudhir-serpentcs: Yes, I do agree with you.
I will remove it.

sudhir-serpentcs:
Removed duplicate images from README and added anonymous function to get default company.

sudhir-serpentcs:
Next: Replace old screenshots in README by new ones for v9.

elicoidal:
True and add some simple demo data as well ( @sudhir-serpentcs )

elicoidal:
@sudhir-serpentcs demo data can be done in second stage. Can you resubmit with newer screenshot so that we have it reviewed and merged?

moylop260 (changes requested):

sudhir-serpentcs:
Replaced old screenshots with new ones, removed string parameters from the fields.

sudhir-serpentcs:
Merged the PR of Astirpe.

business_requirement/README.rst
```diff
@@ -0,0 +1,230 @@
...
+* Trading (New product development)
+* Business Consultancy
+* Web or IT development
+* R&D projects
+* etc.
+
+More information about business requirements management:
+
+* `Wikipedia <https://en.wikipedia.org/wiki/Business_requirements>`_
+* `Six Sigma <(https://www.isixsigma.com/implementation/project-selection-tracking/business-requirements-document-high-level-review/>`_
```
elicoidal: remove "(" in URL

elicoidal (changes requested):

astirpe:
Code review :+1:

business_requirement/models/business_requirement.py
```diff
@@ -0,0 +1,260 @@
...
+        for br in self:
+            br.sub_br_count = len(br.business_requirement_ids)
+
+    @api.model
+    def _get_states(self):
+        states = [
+            ('draft', 'Draft'),
+            ('confirmed', 'Confirmed'),
+            ('approved', 'Approved'),
+            ('stakeholder_approval', 'Stakeholder Approval'),
```
elicoidal: stakeholder_approved, Stakeholder Approved

business_requirement/models/business_requirement.py
```diff
@@ -0,0 +1,260 @@
...
+
+    @api.multi
+    def action_button_approve(self):
+        self.state = 'approved'
+        self.approved_id = self.env.user
+        self.approval_date = fields.Datetime.now()
+
+    @api.multi
+    def action_button_stakeholder_approval(self):
+        self.state = 'stakeholder_approval'
```
elicoidal: modify here as well

business_requirement/models/business_requirement.py
```diff
@@ -0,0 +1,260 @@
...
+    def action_button_drop(self):
+        self.state = 'drop'
+
+    @api.cr_uid_ids_context
+    def message_post(self, cr, uid, thread_id, body='', subject=None,
+                     subtype=None, parent_id=False,
+                     attachments=None, context=None,
+                     content_subtype='html', **kwargs):
+        subject = None
+        if context.get(
```
elicoidal, line 247: formatting is a little bit weird but non blocking

elicoidal (changes requested):
small details. Other than that :+1:

business_requirement/models/business_requirement.py
astirpe: @elicoidal in V8 the states values have `stakeholder_approval`, we should be careful when changing this item. It also implies to create migration scripts.

Maybe changing only the string 'Stakeholder Approved' would be enough?

elicoidal: Agree. Let's just change the name

astirpe: https://github.com/sudhir-serpentcs/business-requirement/pull/9

business_requirement/models/business_requirement.py
astirpe, line 223: Modification not required anymore, changing the name https://github.com/sudhir-serpentcs/business-requirement/pull/9 is enough

dreispt:
@moylop260 @elicoidal Are there still any blocking points?

moylop260 (approved):

elicoidal:
@dreispt sorry for the delay, just catching up.
I think this is safe to merge.

PR #49, opened 2016-10-08
Fix misspelled field name linked_project in view of model

mtelahun:
In the view of module business_requirement_deliverable_project the 'attrs' attribute of generate_projects_wizard button misspells the field linked_project as link_project. This prevents the form view of the model business_requirement from loading.


elicoidal:
:+1: FT

PR #50, opened 2016-10-09
9.0 business requirement add test case field

RawEvan:
This PR adds `test_case` field for business requirement module.
Based on: https://github.com/OCA/business-requirement/pull/25


elicoidal:
better amend #25

PR #53, opened 2016-10-26
[FIX] Saving of Resource lines when type is procurement

### LINKED ISSUE #43: Resource lines under Deliverables lines tab
if the Type = "task" and the Root Category for tasks and Assign to columns have values, then if we change the Type to "Procurement" values of the two columns are removed, But; when save and close the Deliverables lines then reopen it we find that the Root Category for tasks and Assign to columns will remain read only but will show their values as if the Type="task" while they should show nothing and being set to Null.

astirpe:
Fixes #43


elicoidal:
:+1:

susport (approved):
👍

rafaelbn (approved):

PR #58, opened 2016-11-15
Merged victor prs

franksongfeng:
add category for task generated in business requirement.


franksongfeng:
On wrong branch. It should be on v8.0

PR #61, opened 2016-12-13
[FIX] business_requirement_deliverable_cost: Issue found when adding …

martintorresvm:
…resources lines.

	modified:   business_requirement_deliverable_cost/models/business.py

### DIFF
--- a/business_requirement_deliverable_cost/models/business.py
+++ b/business_requirement_deliverable_cost/models/business.py
@@ -46,12 +46,15 @@ class BusinessRequirementResource(models.Model):
     @api.multi
     def _get_partner(self):
         self.ensure_one()
-        partner_id = False
-        if self.business_requirement_deliverable_id.\
-                business_requirement_id.partner_id.id:
-            partner_id = self.business_requirement_deliverable_id.\
-                business_requirement_id.partner_id
-        return partner_id
+        br_id = br_deliverable = False
+        if self.business_requirement_deliverable_id.id:
+            br_deliverable = self.business_requirement_deliverable_id
+        if br_deliverable.business_requirement_id.id:
+            br_id = br_deliverable.business_requirement_id
+        if br_id and br_id.partner_id:
+            return br_id.partner_id
+        else:
+            return False
 
     @api.multi
     def _get_pricelist(self):
--- a/business_requirement_deliverable_cost/models/business.py
+++ b/business_requirement_deliverable_cost/models/business.py
@@ -49,7 +49,8 @@ class BusinessRequirementResource(models.Model):
         br_id = br_deliverable = False
         if self.business_requirement_deliverable_id.id:
             br_deliverable = self.business_requirement_deliverable_id
-        if br_deliverable and br_deliverable.business_requirement_id.id:
+        if self.business_requirement_deliverable_id.\
+                business_requirement_id.partner_id:
             br_id = br_deliverable.business_requirement_id
         if br_id and br_id.partner_id:
             return br_id.partner_id
--- a/business_requirement_deliverable_cost/models/business.py
+++ b/business_requirement_deliverable_cost/models/business.py
@@ -49,7 +49,7 @@ class BusinessRequirementResource(models.Model):
         br_id = br_deliverable = False
         if self.business_requirement_deliverable_id.id:
             br_deliverable = self.business_requirement_deliverable_id
-        if br_deliverable.business_requirement_id.id:
+        if br_deliverable and br_deliverable.business_requirement_id.id:
             br_id = br_deliverable.business_requirement_id
         if br_id and br_id.partner_id:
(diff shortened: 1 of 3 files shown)


martintorresvm:
@dreispt @pedrobaeza @moylop260

pedrobaeza:
It works the same, so no need to change it.

martintorresvm:
> It works the same, so no need to change it.

@pedrobaeza But I have this error when try to add a new resource line:

screenshot-3275181-8-0-dfd030 runbot1 odoo-community org-2016-12-13-15-24-18

pedrobaeza:
If `business_requirement_deliverable_id` is a many2one field, then `business_requirement_deliverable_id.id` will be False, so the error should come from other place.

martintorresvm:
> If business_requirement_deliverable_id is a many2one field, then business_requirement_deliverable_id.id will be False, so the error should come from other place.

@pedrobaeza Thanks you are right. I thought too much but then it is simpler thanks to your notions

business_requirement_deliverable_cost/models/business.py
```diff
@@ -46,11 +46,8 @@ def _compute_sale_price_total(self):
     @api.multi
     def _get_partner(self):
         self.ensure_one()
-        br_id = br_deliverable = False
-        if self.business_requirement_deliverable_id.id:
-            br_deliverable = self.business_requirement_deliverable_id
-        if br_deliverable.business_requirement_id.id:
-            br_id = br_deliverable.business_requirement_id
+        br_id = self.business_requirement_deliverable_id.\
```
pedrobaeza: This is going to fail also on next line, as you are storing the integer ID.

martintorresvm: Now is fixed

martintorresvm:
I am trying to fix coveralls but not sure in which part.
@gurneyalex Could you help?

screenshot-coveralls io-2016-12-13-18-22-50

moylop260:
@victormartinelicocorp better use codecov result because coveralls has a few issues about diff coverage patch

martintorresvm:
@dreispt @pedrobaeza @moylop260 Could be merged this PR?

business_requirement_deliverable_cost/models/business.py
```diff
@@ -49,7 +49,7 @@ def _get_partner(self):
         br_id = br_deliverable = False
         if self.business_requirement_deliverable_id.id:
             br_deliverable = self.business_requirement_deliverable_id
-        if br_deliverable.business_requirement_id.id:
+        if br_deliverable and br_deliverable.business_requirement_id.id:
```
moylop260: What about use `self.business_requirement_deliverable_id.business_requirement_id.partner_id`?

martintorresvm: probably much better, but seriously I couldn't understand 100% how it works Coveralls since I only see a diff on code as an explanation of decreasing coverall so thats why I back to the oldest code. I will commit again with the improvement you marked and let's see if coveralls remains the same.

moylop260: We are using coveralls because historical reason.

But you could see the reasons to use other one here (codecov)

https://github.com/OCA/maintainer-quality-tools/pull/342

martintorresvm:
@moylop260 I modified it

martintorresvm:
@pedrobaeza @moylop260 @dreispt @elicoidal Could be merged?

business_requirement_deliverable_cost/models/business.py
```diff
@@ -49,7 +49,8 @@ def _get_partner(self):
         br_id = br_deliverable = False
         if self.business_requirement_deliverable_id.id:
             br_deliverable = self.business_requirement_deliverable_id
-        if br_deliverable.business_requirement_id.id:
+        if self.business_requirement_deliverable_id.\
```
moylop260: Are you getting the same value 3 times or I didn't understand this code?

I mean, What about use?
```python
     @api.multi
     def _get_partner(self):
         self.ensure_one()
         return self.business_requirement_deliverable_id.business_requirement_id.partner_id
```

moylop260: Note: In fact, I don't know if is better a related

elicoidal:
@victormartinelicocorp please review pending questions and then we could probably merge

martintorresvm:
Here please note that project coverage for "codecov" is ok but for coverage/coveralls IMO decrease by due to the offset of the removed lines (https://docs.codecov.io/docs/coverage-offset)

@pedrobaeza @moylop260 @elicoidal Please have a check if everything is satisfied so we could merge it?

I think there is only one comment pending from @moylop260 I solved on my latest commit.

elicoidal (approved):

moylop260 (approved):

PR #92, opened 2017-03-22
[8.0][FIX] Keep master project when generate project

seb-elico:
When generating a project from a BR, the generated project should be stored in the "Linked Project" (`linked_project`), not in the "Master Project" (`project_id`)

### DIFF
--- a/business_requirement_deliverable_project/models/project.py
+++ b/business_requirement_deliverable_project/models/project.py
@@ -55,7 +55,7 @@ class Project(models.Model):
             br_ids.filtered(lambda br_id: not br_id.parent_id)
         vals = {
             'partner_id': self.partner_id.id,
-            'project_id': self.id,
+            'linked_project': self.id,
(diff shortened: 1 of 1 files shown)


elicoidal (approved):

elicoidal:
passed all checks!
@dreispt @pedrobaeza can you help move forward?

pedrobaeza (approved):

seb-elico:
@elicoidal @victormartinelicocorp @pedrobaeza after testing in Runbot, issue is not fixed. Should we revert this PR or create an additional one to finish the fix?

martintorresvm:
> @elicoidal @victormartinelicocorp @pedrobaeza after testing in Runbot, issue is not fixed. Should we revert this PR or create an additional one to finish the fix?

@seb-elico IMO if not fixed but merged means code that maybe introduce other issue (I mean functional because all test have been passed) on other side. Better revert if we could and continue until it's done

Issue #95, opened 2017-03-24
[8.0] Master project overwritten with generated project


seb-elico:
Current behavior: When generating a project from a BR, the Master Project gets replaced by the generated project.

Expected behavior: Master Project of a BR must not change when generating the project.

I created #92 hoping it could solve the issue but Runbot wasn't working by then. Now I tested it and it doesn't fix the issue, unfortunately the PR has been merged already.

PR #97, opened 2017-03-27
[8.0][FIX] business_requirement_deliverable_project

martintorresvm:
Fix open issue regards:
https://github.com/OCA/business-requirement/issues/95
On relations One2Many need to be careful using structures
like [6, 0, [ids]]

### DIFF
--- a/business_requirement_deliverable_project/wizard/generate_projects.py
+++ b/business_requirement_deliverable_project/wizard/generate_projects.py
@@ -166,9 +166,9 @@ class BrGenerateProjects(models.TransientModel):
             description = br.description
             privacy_visibility = br.project_id.privacy_visibility \
                 or br.project_id._defaults['privacy_visibility']
-            br_id = br.id
-        elif br._name == 'business.requirement.deliverable':
-            br_id = br.business_requirement_id.id
+        #     br_id = br.id
+        # elif br._name == 'business.requirement.deliverable':
+        #     br_id = br.business_requirement_id.id
         vals = {
             'name': description,
             'parent_id': parent.analytic_account_id.id,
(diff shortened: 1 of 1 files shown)


martintorresvm:
@elicoidal @seb-elico

business_requirement_deliverable_project/wizard/generate_projects.py
```diff
@@ -166,9 +166,9 @@ def _prepare_project_vals(self, br, parent):
             description = br.description
             privacy_visibility = br.project_id.privacy_visibility \
                 or br.project_id._defaults['privacy_visibility']
-            br_id = br.id
-        elif br._name == 'business.requirement.deliverable':
-            br_id = br.business_requirement_id.id
+        #     br_id = br.id
```
elicoidal: @victormartinelicocorp 
Your PR looks strange: you comment removed lines.
I would say: remove dead code, do not comment.
Something I missed?

martintorresvm: @elicoidal no you are not missing a thing, sorry but this is dead code. Already I commit to fix it

elicoidal (approved):

elicoidal:
@pedrobaeza @dreispt I think this one is good to go

pedrobaeza:
I don't get the problem that this PR solves.

martintorresvm:
> I don't get the problem that this PR solves.

@pedrobaeza This is a functional problem.

We like to keep the `project_id` as it's created the BR and the previous code overwrite with `linked_project` when generates project on the wizard.

cc @elicoidal

pedrobaeza:
But this one does nothing with linked_project, but with field br_ids...

martintorresvm:
@pedrobaeza It's a relation is one2many on `project.project` so the real `id` is on BR object.

When create a project if I pass the values `br_ids` mean that I will overwrite the BR `project_id` value with the `id` of the project I am creating. 

So indirectly when create the linked project I am replacing the id of project_id on the BR and later I assign to `linked_project` the value which is what is real intend to do.

pedrobaeza (approved):
OK, thanks for the explanations.

Issue #91, opened 2017-03-22
[8.0] improve ACL for sales estimation


elicoidal:
Current behavior
Group "Business Requirement Estimation" is defined in module business-requirement/business_requirement_deliverable_cost/

Expected behavior:

Group "Business Requirement Estimation" is defined in module business-requirement/business_requirement_deliverable/

In tree view following Columns should only be visible for this group in the model and all forms (tree/form):
* unit_price
* price_total
* currency_id

martintorresvm:
@elicoidal the PR https://github.com/OCA/business-requirement/pull/100 regards this issue is almost ready to merge just small changes.

PR #100, opened 2017-03-29
[ADD] Added security groups (Issue #91)

sudhir-serpentcs:
@elicoidal Please review


elicoidal (approved):

elicoidal:
@pedrobaeza @dreispt thanks for your help

business_requirement_deliverable/views/business_view.xml
```diff
@@ -31,14 +31,14 @@
             <field name="model">business.requirement.deliverable</field>
             <field name="arch" type="xml">
                 <tree string="Business Requirement Deliverable" default_order="sequence">
-                    <field name="currency_id" invisible="1"/>
+                    <field name="currency_id" invisible="1" groups="business_requirement_deliverable.group_business_requirement_estimation"/>
```
pedrobaeza: Why an invisible field required groups?

business_requirement_deliverable/views/business_view.xml
```diff
@@ -68,15 +68,15 @@
                 <form string="Business Requirement Task Type">
                     <group>
                         <group>
-                            <field name="currency_id" invisible="1"/>
+                            <field name="currency_id" invisible="1" groups="business_requirement_deliverable.group_business_requirement_estimation"/>
```
pedrobaeza: The same

pedrobaeza (approved):
Little nitpicking

business_requirement_deliverable/views/business_view.xml
elicoidal: @pedrobaeza There are 3 levels of information with different levels of ACL in the BR:
* Functional: visible by the BR user
* Sales: visible by estimation group
* Cost: visible by the cost group

martintorresvm: @elicoidal I think @pedrobaeza is right

The meaning to add parameter `groups` into the view is to make this field only visible for those groups listed is not like a real ACL.

So actually I would ask @sudhir-serpentcs to remove parameter invisible so now will make sense.

business_requirement_deliverable/views/business_view.xml
martintorresvm: @sudhir-serpentcs please remove invisible parameter.

business_requirement_deliverable/views/business_view.xml
elicoidal: @victormartinelicocorp 
What I want is https://github.com/OCA/business-requirement/pull/100#discussion_r108822624
I trust that you will find the right method to do it.

martintorresvm: @elicoidal this field mean for mean to by only visible on Cost level users, right?

elicoidal: only visible to estimate users

sudhir-serpentcs:
victormartinelicocorp 2 hours ago
    @sudhir-serpentcs please remove invisible parameter.
@victormartinelicocorp @elicoidal @pedrobaeza Done.

elicoidal:
LGTM. could be merge IMO @pedrobaeza ?

PR #103, opened 2017-04-01
[8.0][IMP] Changed Workflow for business requirements.

sudhir-serpentcs:
Changed workflow.
Here's the matrix to summarize (V = editable, X = read-only):
                       Draft  Confirmed  Approved  SH Appr.  In prog.
Cust. Story     V               X                      X             X              X
Scenario         V                V                     X              X             X
Gap                  V               V                      X              X             X
QA                    V               V                     X               X             X
Deliverables   V              V                    X                X             X
Resources       V              V                    X                X             X
Assignee        V               V                    V                 V             X

@elicoidal

PR #113, opened 2017-04-10
[IMP][BR_deliverable] Added Smart Buttons and Search views for DL and RL

sudhir-serpentcs:

Added Smart buttons for Resource Lines and Deliverble Lines  also added Search views

### DIFF
--- a/business_requirement_deliverable/models/business.py
+++ b/business_requirement_deliverable/models/business.py
@@ -5,6 +5,7 @@ from openerp import api, fields, models
 from openerp.exceptions import Warning as UserError
 from openerp.exceptions import ValidationError
 from openerp.tools.translate import _
+from lxml import etree
 
 
 class BusinessRequirementResource(models.Model):
@@ -66,6 +67,23 @@ class BusinessRequirementResource(models.Model):
         string='Business Requirement',
         store=True
     )
+
+    @api.model
+    def fields_view_get(self, view_id=None,
+                        view_type='form', toolbar=False, submenu=False):
+        result = super(BusinessRequirementResource,
+                       self).fields_view_get(view_id, view_type,
+                                             toolbar=toolbar, submenu=submenu)
+        if view_type in ['form', 'tree']:
+            business_requirement = self.env['business.requirement'].browse(
+                self._context.get('active_id')).state
+            if business_requirement not in ['draft', 'confirmed']:
+                doc = etree.XML(result['arch'])
+                nodes = doc.xpath("//tree")
+                for node in nodes:
+                    node.set('edit', '0')
+                result['arch'] = etree.tostring(doc)
+        return result
 
     @api.multi
     @api.onchange('product_id')
@@ -163,6 +181,23 @@ class BusinessRequirementDeliverable(models.Model):
         string='Business Requirement',
         store=True
     )
+
+    @api.model
+    def fields_view_get(self, view_id=None,
+                        view_type='form', toolbar=False, submenu=False):
+        result = super(BusinessRequirementDeliverable,
+                       self).fields_view_get(view_id, view_type,
+                                             toolbar=toolbar, submenu=submenu)
+        if view_type in ['form', 'tree']:
+            business_requirement = self.env['business.requirement'].browse(
+                self._context.get('active_id')).state
+            if business_requirement not in ['draft', 'confirmed']:
+                doc = etree.XML(result['arch'])
+                nodes = doc.xpath("//form")
+                for node in nodes:
+                    node.set('edit', '0')
+                result['arch'] = etree.tostring(doc)
+        return result
 
     @api.multi
     @api.depends('business_requirement_id.partner_id')
--- a/business_requirement_deliverable/models/business.py
+++ b/business_requirement_deliverable/models/business.py
@@ -5,22 +5,11 @@ from openerp import api, fields, models
 from openerp.exceptions import Warning as UserError
 from openerp.exceptions import ValidationError
 from openerp.tools.translate import _
-from lxml import etree
 
 
 class BusinessRequirementResource(models.Model):
     _name = "business.requirement.resource"
     _description = "Business Requirement Resource"
-
-    @api.model
-    def default_get(self, fields):
-        res = super(BusinessRequirementResource, self).default_get(fields)
-        if self._context.get('business_requirement_id'):
-            res.update({
-                'business_requirement_id':
-                self._context.get('business_requirement_id')
-            })
-        return res
 
     sequence = fields.Integer('Sequence')
     name = fields.Char('Name', required=True)
@@ -77,23 +66,6 @@ class BusinessRequirementResource(models.Model):
         string='Business Requirement',
         store=True
     )
-
-    @api.model
-    def fields_view_get(self, view_id=None,
-                        view_type='form', toolbar=False, submenu=False):
-        result = super(BusinessRequirementResource,
-                       self).fields_view_get(view_id, view_type,
-                                             toolbar=toolbar, submenu=submenu)
-        if view_type in ['form', 'tree']:
-            business_requirement = self.env['business.requirement'].browse(
-                self._context.get('active_id')).state
-            if business_requirement not in ['draft', 'confirmed']:
-                doc = etree.XML(result['arch'])
-                nodes = doc.xpath("//tree")
-                for node in nodes:
-                    node.set('edit', '0')
-                result['arch'] = etree.tostring(doc)
-        return result
 
     @api.multi
     @api.onchange('product_id')
@@ -163,7 +135,8 @@ class BusinessRequirementDeliverable(models.Model):
     business_requirement_id = fields.Many2one(
         comodel_name='business.requirement',
         string='Business Requirement',
-        ondelete='cascade'
+        ondelete='cascade',
+        required=True
     )
     unit_price = fields.Float(
         string='Sales Price'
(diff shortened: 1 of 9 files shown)


elicoidal:
Please fix travis

elicoidal:
@sudhir-serpentcs 

Here is the result of my initial functional tests:
- [x] You should remove the tab "Deliverable" from BR
- [x] The access rules from #103 should still be valid
- [x] Smart buttons should indicate how many RL and DL are available

In DL tree view:
- [x] remove Currency
- [x] Add BR (in place of currency)

In DL Form View
- [x] Currency should be after the total
- [x] BR field should be before the product
- [x] Have 2 columns: 
  * first one with BR+Product+Qty+unit+sales Price+total+currency
  * second: Name
- [ ] In RL sub form: remove the root task category field ( @victormartinelicocorp not sure about the usage)
- [x] In RL sub form: hide the BR field

In RL Tree view
- [x] Move the BR in first place
- [x] Add the DL after the BR

No tests on prices yet.

sudhir-serpentcs:
please review code

elicoidal:
@sudhir-serpentcs thanks for the update, getting good.
I have updated comment https://github.com/OCA/business-requirement/pull/113#issuecomment-292891450 with progress.

Besides here are some new finding:
RL  view:
- [X] In RL tree view: should not see the task category field (To be ignored as the module will be removed later @victormartinelicocorp )
- [X] in RL tree view: add the graph/pivot view
- [X] In RL tree view, in the field, only the deliverable lines from the selected BR should appear
- [X] In RL tree view you should not be able to select a BR if state is not draft/confirmed.

DL view
- [X] in DL tree view: add the graph/pivot view
- [X] In DL form view  you should not be able to select a BR if state is not draft/confirmed.
- [X] In DL, BR should be mandatory
- [X] In DL, I cannot see anymore the cost and prices at RL level
- [X] In DL, I cannot see anymore the Sales and prices at RL level
~In DL, I cannot see the fields Category (but this is maybe due to inheritence)~

Misc:
- [X] Add menu entry for Deliverable lines and Resources lines below BR entry

business_requirement_deliverable/models/business.py
```diff
@@ -5,12 +5,23 @@
...
 from openerp.tools.translate import _
+from lxml import etree
 
 
 class BusinessRequirementResource(models.Model):
     _name = "business.requirement.resource"
     _description = "Business Requirement Resource"
 
+    @api.model
+    def default_get(self, fields):
```
martintorresvm: @sudhir-serpentcs Not blocking but could you explain we is need to modify `defaul_get`?

business_requirement_deliverable/models/business.py
```diff
@@ -49,6 +60,40 @@ class BusinessRequirementResource(models.Model):
...
+    )
+    business_requirement_project_id = fields.Many2one(
+        comodel_name='project.project',
+        related='business_requirement_id.project_id',
+        string='Business Requirement',
+        store=True
+    )
+
+    @api.model
+    def fields_view_get(self, view_id=None,
```
martintorresvm: @sudhir-serpentcs The same for this  `fields_view_get`

sudhir-serpentcs: @victormartinelicocorp 

access rules for #103

business_requirement_deliverable/models/business.py
sudhir-serpentcs: @victormartinelicocorp 
will be removed from next commit

business_requirement_deliverable/models/business.py
martintorresvm: @sudhir-serpentcs Please do the more specif you can the PR. I wish to see that on separated PR and for the proper module.

And by they way this could be done IMO simpler like is actually done:

Draft Confirmed Approved SH Appr. In prog
Cust. Story V X X X X
Actually works as the requirement see the `states` param
```
    business_requirement = fields.Html(
        'Customer Story',
        readonly=True,
        states={'draft': [('readonly', False)]}
    )

```

Do still think need to overwrite `fields_view_get` ?
I would call it editable rules instead of access rules

cc @elicoidal

martintorresvm (changes requested):

business_requirement_deliverable/models/business.py
martintorresvm: @sudhir-serpentcs Sorry but my comment for the unnecessary explanation. 

I want to reformulate my question this it's done #103 
What is exactly is fixed by overwrite the function?

sudhir-serpentcs:
@victormartinelicocorp  I do agree with you but state field is not in the object. We need to make RL / DL R/O if BR is not in draft / confirmed state.

martintorresvm (approved):
@sudhir-serpentcs Thanks for the explanation

sudhir-serpentcs:
@victormartinelicocorp  @elicoidal 
please review code

@victormartinelicocorp  by mistake fields_view_get removed from #ebe0f96 
so i have revised on next commit #e7f3596

elicoidal:
@sudhir-serpentcs I have made some tests but only few has been updated. Check https://github.com/OCA/business-requirement/pull/113#issuecomment-293487176

sudhir-serpentcs:
@elicoidal 

 #113 (comment)In DL, I cannot see anymore the cost and prices at RL level + cannot see the fields Category
:-  you mean that i have to add cost field and price total in RL 

test1

Misc:

 Add menu entry for Deliverable lines and Resources lines below BR entry
you mean that i have to add menu after Business Requirment

test3

elicoidal:
> you mean that i have to add cost field and price total in RL

I think so. I see them in RL but not in DL/RL. it should be similar.

> you mean that i have to add menu after Business Requirment

Correct

Thanks!

sudhir-serpentcs:

please review code

elicoidal:
@sudhir-serpentcs It is getting close to merge. I have added one topic in https://github.com/OCA/business-requirement/pull/113#issuecomment-293487176 about sales prices in RL

The rest is good. Once fixed we can start merging and adapt other modules to the new structure.

sudhir-serpentcs:

please review code

elicoidal (approved):
:+1:

elicoidal:
@victormartinelicocorp @seb-elico please review

elicoidal:
@pedrobaeza @dreispt @seb-elico Let's move forward?

business_requirement_deliverable/models/business.py
```diff
@@ -49,6 +50,40 @@ class BusinessRequirementResource(models.Model):
...
+                        view_type='form', toolbar=False, submenu=False):
+        result = super(BusinessRequirementResource,
```
pedrobaeza: Add docstring saying for what is this needed. Reading code, I don't see this very clear and with possible side effects, making editable a tree based on the active_id (which I suppose that can be changed navigating through records).

business_requirement_deliverable/models/business.py
```diff
@@ -133,6 +169,35 @@ class BusinessRequirementDeliverable(models.Model):
...
+        store=True
+    )
+
+    @api.model
+    def fields_view_get(self, view_id=None,
+                        view_type='form', toolbar=False, submenu=False):
+        result = super(BusinessRequirementDeliverable,
+                       self).fields_view_get(view_id, view_type,
+                                             toolbar=toolbar, submenu=submenu)
+        if view_type in ['form', 'tree']:
```
pedrobaeza: The same comment.

business_requirement_deliverable/models/business.py
```diff
@@ -133,6 +169,35 @@ class BusinessRequirementDeliverable(models.Model):
...
+            business_requirement = self.env['business.requirement'].browse(
+                self._context.get('active_id')).state
+            if business_requirement not in ['draft', 'confirmed']:
+                doc = etree.XML(result['arch'])
+                nodes = doc.xpath("//form")
```
pedrobaeza: Form can be non editable this way?

business_requirement_deliverable/models/business.py
```diff
@@ -244,6 +309,68 @@ class BusinessRequirement(models.Model):
...
+
+    @api.multi
+    def _compute_dl_count(self):
+        self.dl_count = self.env['business.requirement.deliverable'].\
+            search_count([('business_requirement_id', '=', self.id)])
+
+    @api.multi
+    def _compute_rl_count(self):
+        for r in self:
+            r.rl_count = self.env['business.requirement.resource'
```
pedrobaeza: Please use another indentation scheme that is not so ugly. Example:
```
r.rl_count = self.env[
    'business.requirement.resource'
].search_count(
    [('business_requirement_id', '=', r.id)
)
```
Anyway this can be replace by a one2many field, and use `len()` over that field.

pedrobaeza (changes requested):

elicoidal:
@pedrobaeza thanks for the comments!

sudhir-serpentcs:
@pedrobaeza thanks for the suggestions. We will work on that and update the PR soon.

sudhir-serpentcs:
@pedrobaeza 

Changed the code as per comment.

CC @elicoidal

elicoidal:
@pedrobaeza can we move forward?

pedrobaeza:
I don't see any docstring documenting fields_view_get methods nor my questions answer, as I doubt about the feasibility of the method applied to both form and tree.

elicoidal:
@pedrobaeza you are right: I missed it.
@sudhir-serpentcs let us know when you can fix the remaining comments?

sudhir-serpentcs:
@pedrobaeza

Please review it.

CC @elicoidal

business_requirement_deliverable/models/business.py
```diff
@@ -49,6 +50,46 @@ class BusinessRequirementResource(models.Model):
...
+        if view_type in ['form', 'tree']:
+            business_requirement = self.env['business.requirement'].\
+                browse(self._context.get('active_id')).state
+            if business_requirement not in ['draft', 'confirmed']:
+                doc = etree.XML(result['arch'])
+                nodes = doc.xpath("//tree")
+                for node in nodes:
+                    node.set('edit', '0')
+                result['arch'] = etree.tostring(doc)
+        return result
```
dreispt: I'm not in favour of this type of hacking.
It would be both simpler and customization-friendly to enforce this using Record Rules based on State.

elicoidal: OK

sudhir-serpentcs:
@dreispt 

Please review code.

CC @elicoidal

elicoidal:
@pedrobaeza @dreispt let's know whether we can move on

pedrobaeza (approved):

pedrobaeza:
Please squash by author and logical set before merging

Issue #114, opened 2017-04-11
BR def name_get() Issue when use Group By on Task by business-requirement


martintorresvm:
BR name_get use format() built in Python lib and if some string is unicode will happen like this:

Apr 11 10:44:25 superman docker/trunk_odoo-elico_1[806]:   File "/opt/odoo/additional_addons/business-requirement/business_requirement/models/business.py", line 256, in name_get
Apr 11 10:44:25 superman docker/trunk_odoo-elico_1[806]:     formatted_name = '[{}] {}'.format(br.name, br.description)
Apr 11 10:44:25 superman docker/trunk_odoo-elico_1[806]: UnicodeEncodeError: 'ascii' codec can't encode characters in position 29-37: ordinal not in range(128

martintorresvm:
@elicoidal @seb-elico

seb-elico:
@victormartinelicocorp maybe you can try with the `%` substitution

martintorresvm:
Yes indeed it's need to be changed by another method. I will create an Issue as well for us. 
Here I need @elicoidal add the tags `bug` and `help wanted`

PR #116, opened 2017-04-13
[FIX][business_requirement] Name formating improvement

martintorresvm:
To fix https://github.com/OCA/business-requirement/issues/114


martintorresvm:
@elicoidal @seb-elico

cc @sudhir-serpentcs

elicoidal (approved):

elicoidal:
@pedrobaeza @dreispt @seb-elico Let's move forward?

business_requirement/models/business.py
```diff
@@ -251,9 +251,11 @@ def name_get(self):
         result = []
         for br in self:
             if br.ref:
-                formatted_name = '[{}] {}'.format(br.ref, br.description)
```
pedrobaeza, line 262: The solution is to put:
```
formatted_name = u'[{}] {}'.format(br.ref, br.description)
```
not to make the other change

pedrobaeza (changes requested):

business_requirement/models/business.py
martintorresvm: Sure, thanks for review!

martintorresvm: done

elicoidal:
@pedrobaeza your comment is attended so I move forward :). Thanks for the review!

PR #123, opened 2017-05-03
[ADD] business_requirement_deliverable_resource_template

martintorresvm:


martintorresvm:
@elicoidal @seb-elico Please have a check. Also I'll push asap one FIX for BRD module regards this one.

business_requirement_deliverable_resource_template/models/business_requirement_resource_template.py
```diff
@@ -0,0 +1,84 @@
...
+from openerp.exceptions import ValidationError
+from openerp.tools.translate import _
+
+
+class BusinessRequirementDeliverable(models.Model):
+    _inherit = "business.requirement.deliverable"
+
+    def _prepare_resource_lines(self):
+        rl_data = self.product_id.sudo().resource_lines.copy_data()
+        rl_data = [(0, 0, item) for index, item in enumerate(rl_data)]
```
seb-elico: @victormartinelicocorp you can return this line directly

business_requirement_deliverable_resource_template/models/business_requirement_resource_template.py
```diff
@@ -0,0 +1,84 @@
...
+        product = self.product_id
+        if product:
+            self.resource_ids = self._prepare_resource_lines()
+        business_requirement = self.business_requirement_id
+        if business_requirement:
+            for resource in self.resource_ids:
+                resource.business_requirement_id = business_requirement.id
+
+
+class BusinessRequirementResource(models.Model):
```
seb-elico: Shouldn't it be `BusinessRequirementResourceTemplate`?

martintorresvm:
@elicoidal @seb-elico  We haven't define what should happens with ACL for new model resources templates.

br user I give readonly access
brd estimation I give full access

seb-elico (changes requested):

elicoidal:
Agree

elicoidal (approved):

business_requirement_deliverable_resource_template/models/business_requirement_resource_template.py
martintorresvm: @seb-elico Weird but that causes E901 SyntaxError: invalid syntax on my Lint checker. Good point anyway.

martintorresvm:
@seb-elico Thanks for your review

business_requirement_deliverable_resource_template/models/business_requirement_resource_template.py
```diff
@@ -0,0 +1,83 @@
...
+    @api.multi
+    @api.constrains('resource_type', 'uom_id')
+    def _check_description(self):
+        for resource in self:
+            if resource.resource_type == 'task' and (
+                    resource.uom_id.category_id != (
+                        self.env.ref('product.uom_categ_wtime'))):
+                raise ValidationError(_(
+                    "When resource type is task, "
+                    "the uom category should be time"))
```
seb-elico: s/uom/UoM/

seb-elico (changes requested):

martintorresvm:
@dreispt @pedrobaeza Could you review this PR?

I will create a new PR to deprecate the module `business_requirement_deliverable_default` since now `business_requirement_id` is required on RL it's really hard to have RL templates that you should preassign to a BR.

cc @elicoidal @seb-elico

seb-elico (approved):
LGTM

business_requirement_deliverable_resource_template/views/business_requirement_deliverable_default.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+    <data>
+        <record id="product_template_only_form_view" model="ir.ui.view">
+            <field name="name">product.template.form</field>
+            <field name="model">product.template</field>
+            <field name="inherit_id" ref="product.product_template_only_form_view"/>
+            <field name="arch" type="xml">
+                <page name="variants" position='after'>
+                    <page name="resources" string="Default Resource lines">
```
dreispt: "Lines" in uppercase

business_requirement_deliverable_resource_template/models/business_requirement_resource_template.py
```diff
@@ -0,0 +1,83 @@
...
+from openerp.exceptions import ValidationError
+from openerp.tools.translate import _
+
+
+class BusinessRequirementDeliverable(models.Model):
+    _inherit = "business.requirement.deliverable"
+
+    def _prepare_resource_lines(self):
+        rl_data = self.product_id.sudo().resource_lines.copy_data()
+        return [(0, 0, item) for index, item in enumerate(rl_data)]
```
dreispt, line 14: This method should return a recordset.

business_requirement_deliverable_resource_template/models/business_requirement_resource_template.py
```diff
@@ -0,0 +1,83 @@
...
+    @api.onchange('product_id')
+    def product_id_change(self):
+        super(BusinessRequirementDeliverable, self).product_id_change()
+        product = self.product_id
+        if product:
+            self.resource_ids = self._prepare_resource_lines()
+        business_requirement = self.business_requirement_id
+        if business_requirement:
+            for resource in self.resource_ids:
+                resource.business_requirement_id = business_requirement.id
```
dreispt: In the new API you work with record objects, not `id` numbers: drop the `.id`

business_requirement_deliverable_resource_template/models/business_requirement_resource_template.py
```diff
@@ -0,0 +1,83 @@
...
+        comodel_name='product.uom',
+        string='UoM',
+        required=True
+    )
+    qty = fields.Float(
+        string='Quantity',
+        default=1,
+    )
+    resource_type = fields.Selection(
+        selection=[('task', 'Task'), ('procurement', 'Procurement')],
```
dreispt: This selection list items would be best defined as a constant value in `business_requirement_deliverable`, then imported and reused here.
The point is, if in the future we want to add more types, we should do the change only in one place.

business_requirement_deliverable_resource_template/models/business_requirement_resource_template.py
```diff
@@ -0,0 +1,83 @@
...
+    @api.multi
+    @api.constrains('resource_type', 'uom_id')
+    def _check_description(self):
+        for resource in self:
+            if resource.resource_type == 'task' and (
+                    resource.uom_id.category_id != (
+                        self.env.ref('product.uom_categ_wtime'))):
+                raise ValidationError(_(
+                    "When resource type is task, "
+                    "the UoM category should be time"))
```
dreispt: From an app architecture PoV, it makes sense a "resource" should have a list of ellegible "UoM", and the UoM selection list should be limited to the configured choices.
Maybe this is an idea for a later improvement?

dreispt (commented):
LGTM, some comments to fix.

business_requirement_deliverable_resource_template/models/business_requirement_resource_template.py
martintorresvm: @dreispt Do you mean to create a new object for resource types? Create with initial data and let the user decide if want to add or remove?. Maybe also add a boolean for the resource type that suits for task?

If that this change will impact the BRD_project module as well.

@elicoidal ?

dreispt: No. I mean create a constant variable in on place to be reused in multiple modules.
The future extension I'm thinking about is a customization module, not a UI table.

business_requirement_deliverable_resource_template/models/business_requirement_resource_template.py
martintorresvm: @dreispt Ideally this constrain or validation error should not happen

For a later improvement I think we should restrict UoM list to the `product_id` UoM category.

And if the product category is `Time` should only possible choose resource type Task

Could be done later?

business_requirement_deliverable_resource_template/models/business_requirement_resource_template.py
martintorresvm: @dreispt How to do it on the equivalent number of lines?
With copy_data() I have a dictionary like:
```
[{'name': u'Odoo developer',
  'product_id': 60,
  'qty': 1.0,
  'resource_type': u'task',
  'uom_id': 5},
 {'name': u'Odoo Functional Consultant',
  'product_id': 61,
  'qty': 1.0,
  'resource_type': u'task',
  'uom_id': 5}]
```

then on the return the values prepared are like this:

```
[(0,
  0,
  {'name': u'Odoo developer',
   'product_id': 60,
   'qty': 1.0,
   'resource_type': u'task',
   'uom_id': 5}),
 (0,
  0,
  {'name': u'Odoo Functional Consultant',
   'product_id': 61,
   'qty': 1.0,
   'resource_type': u'task',
   'uom_id': 5})]
```

and this structure of data suit perfect on assignment:

```
self.resource_ids = self._prepare_resource_lines()
```

business_requirement_deliverable_resource_template/models/business_requirement_resource_template.py
elicoidal: Agree with @dreispt but it looks to me like a nice to have.
Can we add a FIXME in the code for later? Would that be OK?

business_requirement_deliverable_resource_template/models/business_requirement_resource_template.py
elicoidal: @dreispt I agree with the idea but at the same time, I couldnot think of another case/distinction and would like to avoid extra complication at that point.
Could we add a FIXME in the code for later reference?

business_requirement_deliverable_resource_template/models/business_requirement_resource_template.py
dreispt: It's just more two lines of code, but not an issue, a FIXME note is fine.

elicoidal:
@dreispt thanks for the review and valuable input

business_requirement_deliverable_resource_template/models/business_requirement_resource_template.py
dreispt: Yes, I already suggested that as a later improvement.

business_requirement_deliverable_resource_template/models/business_requirement_resource_template.py
martintorresvm: @dreispt I tested returning a `recordset` and the result is that is that the new RL are not saved on the DL so IMO doesn't work the same.

If non an issue could be as is it now?

dreispt: I don't see `copy_data()`in the ORM documentation. 
Maybe you should be using [`copy()`](https://github.com/odoo/odoo/blob/10.0/odoo/models.py#L4377) instead? That one is expected to return a recordset.

martintorresvm:
@dreispt Just one comments is pending to solve one but IMO is non and issue 

cc @elicoidal @seb-elico

dreispt (approved):
Just give a try at replacing `copy_date()` with `copy()`.
If you can't make that work, I'm OK to go ahead with the merge.

martintorresvm:
@dreispt Thanks for your point but and even if it's merged I want to reply you

We use that because what I want is not to copy() a record of the same kind. I want to generate a dictionary with the data that I need to create the records for another model.

So I have one product.template with one or more RL.templates and I want copy the information contained
And then generate the RL I expect on the DL.

We spent a quite time on this solution and I think works really fine.

Here is documented the function not sure if need to be on the Odoo Web page documentation to use it
https://github.com/odoo/odoo/blob/8.0/openerp/models.py#L4731

Issue #87, opened 2017-03-16
[business_requirement_deliverable_cost] ACL Update Pricelist button on DL


martintorresvm:
Group:  "business_requirement_deliverable.group_business_requirement_estimation" should be able see button ""Update Prices on Resource Lines"

Also make sure could execute the action linked. ACL or sudo() just on the function 
def action_button_update_estimation(self):

martintorresvm:
@elicoidal

elicoidal:
related to #94

martintorresvm:
> related to #94
@elicoidal this PR is Closed 

@sudhir-serpentcs Could you refer this Issue with the new PR when you have it?

elicoidal:
Fixed by #124

PR #124, opened 2017-05-03
[FIX] issue #87 ACL on Estimation group

serpentcs-dev1:
@victormartinelicocorp Please review.

This PR fixes the group issue in the #87


elicoidal (approved):
FT

PR #129, opened 2017-05-05
[FIX]Fix the Issue on BRD def _compute_rl_total_cost

serpentcs-dev1:
Move field unit_price from BRDCost to BRD

### DIFF
--- a/business_requirement_deliverable/models/business.py
+++ b/business_requirement_deliverable/models/business.py
@@ -66,6 +66,7 @@ class BusinessRequirementResource(models.Model):
         string='Business Requirement',
         store=True
     )
+    unit_price = fields.Float(string='Cost Price')
 
     @api.multi
     @api.onchange('product_id')
@@ -80,6 +81,7 @@ class BusinessRequirementResource(models.Model):
             description += '\n' + product.description_sale
         self.name = description
         self.uom_id = uom_id
+        self.unit_price = self.product_id.standard_price
 
     @api.onchange('resource_type')
     def resource_type_change(self):
@@ -318,7 +320,7 @@ class BusinessRequirement(models.Model):
     def _compute_rl_total_cost(self):
         for r in self:
             for dl in r.deliverable_lines:
-                r.rl_total_cost += sum(rl.price_total for rl in
+                r.rl_total_cost += sum(rl.unit_price for rl in
                                        dl.resource_ids)
 
     @api.multi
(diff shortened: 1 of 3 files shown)


business_requirement_deliverable_cost/views/business.xml
```diff
@@ -52,7 +52,6 @@
                     <field name="gross_profit" groups="business_requirement_deliverable_cost.group_business_requirement_cost_control"/>
                 </xpath>
                 <xpath expr="//field[@name='resource_ids']/tree/field[@name='uom_id']" position="after">
-                    <field name="unit_price" />
```
martintorresvm: @darshan-serpent LGTM but this change please on a different PR

martintorresvm (changes requested):

business_requirement_deliverable/models/business.py
```diff
@@ -66,6 +66,7 @@ class BusinessRequirementResource(models.Model):
         string='Business Requirement',
         store=True
     )
+    unit_price = fields.Float(string='Cost Price')
```
martintorresvm: @darshan-serpent Could you add `groups` parameter like we had on BRDCost?

I think should be add this group
https://github.com/OCA/business-requirement/blob/8.0/business_requirement_deliverable/security/business_requirement_deliverable_security.xml#L5

What do you think @elicoidal ?

martintorresvm (changes requested):

business_requirement_deliverable_cost/views/business.xml
serpentcs-dev1: Okay

elicoidal:
This is not necessary

business_requirement_deliverable/models/business.py
```diff
@@ -66,6 +66,11 @@ class BusinessRequirementResource(models.Model):
         string='Business Requirement',
         store=True
     )
+    unit_price = fields.Float(
```
martintorresvm, line 69: @elicoidal This was fine because is on the Resource model

business_requirement_deliverable/views/business_view.xml
```diff
@@ -350,6 +350,7 @@
                             <field name="user_id" attrs="{'readonly':[('resource_type','=','procurement')]}"/>
                             <field name="qty"/>
                             <field name="uom_id"/>
+                            <field name="unit_price"/>
```
martintorresvm, line 353: @elicoidal And this was fine as well

elicoidal:
Closed by mistake

martintorresvm:
Need review

PR #130, opened 2017-05-05
[Fix] the Issue on BRD_COST def _compute_rl_total_cost

serpentcs-dev1:
Removed field from BRDCost unit_price.


elicoidal:
I think one is not needed to e remoed.

Issue #128, opened 2017-05-04
[8.0] 'business.requirement.resource' object has no attribute 'price_total'


rubencabrera:
The price_total field is (now) defined in the module business_requirement_deliverable_cost, which is not a dependency of business_requirement_deliverable, [where the field is referenced](https://github.com/OCA/business-requirement/blob/8.0/business_requirement_deliverable/models/business.py#L321).

martintorresvm:
Indeed is an issue. We need to decide if remove from calculation or add on **business_requirement_deliverable**

cc @elicoidal @seb-elico

elicoidal:
This is indeed not correct. It should be the sum of the costs in this part of the code.
To be fixed.
@rubencabrera thanks for pointing out!

elicoidal:
@rubencabrera 
We are stabilizing the new version of the modules and somehow the cost and revenue fields got messed up: it will be fixed on Monday.
We are doing now integration testing with our current database (with 1000 Business Requirement database) so we will be able to spot some more in the coming days.
All in all I hope that by end of next week everything is (finally stable)

elicoidal:
Fixed by #131

PR #131, opened 2017-05-05
[FIX] [business_requirement_deliverable] Price fields naming

martintorresvm:
- Fix the meaning of internal field level later will have same name for different meaning
  First means Sales price on one module and later is meaning Cost
- Fix the wrong module to achieve the requirement

### DIFF
--- a/business_requirement_deliverable/models/business.py
+++ b/business_requirement_deliverable/models/business.py
@@ -138,8 +138,9 @@ class BusinessRequirementDeliverable(models.Model):
         ondelete='cascade',
         required=True
     )
-    unit_price = fields.Float(
-        string='Sales Price'
+    sale_price_unit = fields.Float(
+        string='Sales Price',
+        oldname='unit_price'
     )
     price_total = fields.Float(
         compute='_compute_get_price_total',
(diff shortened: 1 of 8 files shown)


martintorresvm:
Also this should fix the Issue https://github.com/OCA/business-requirement/issues/128

Once test are green I'll ask for review

cc @elicoidal @seb-elico

elicoidal (approved):
Fast tracking

elicoidal:
Please fix travis and we are good to go

elicoidal:
Travis is still failing: I assume your fix is not yet complete

business_requirement_deliverable/models/business.py
```diff
@@ -138,8 +138,9 @@ class BusinessRequirementDeliverable(models.Model):
         ondelete='cascade',
         required=True
     )
-    unit_price = fields.Float(
-        string='Sales Price'
+    sale_price_unit = fields.Float(
```
seb-elico, line 141: I think `sales_unit_price` makes more sense.
@elicoidal what do you think?

serpentcs-dev1:
@victormartinelicocorp I guess the travis are failing [here](https://github.com/OCA/business-requirement/blob/8.0/business_requirement_deliverable_cost/views/business.xml#L68)

business_requirement_deliverable/models/business.py
elicoidal: I would like to avoid unnecessary migration and would rather stick to original naming. Not ideal but good enough

martintorresvm:
@elicoidal Travis, Runbot, Codecov/project is Green. 

The others test doesn't mean much is just a measure of total quantity of code cover not the quality of coverage of the project itself

elicoidal:
OK Let's move forward

Issue #127, opened 2017-05-04
[8.0] Onchange product_id in deliverable.


rubencabrera:
This onchange might be a bit annoying when we come with context from another model and are creating a deliverable on the fly. If, from context, we had a description, and select a product, it replaces our text. If we change opinion and delete the selected product, we end up with an empty description, which is even more inconvenient, IMHO.

Is it ok if I create a PR to change this behavior? Any comments on anything I haven't thought of?

martintorresvm:
onchange the product then should place the description only if that moment is empty.

Is what you mean?

rubencabrera:
That would be what I'd expect, are you ok with it?

elicoidal:
@rubencabrera thanks for the proposal: looks like a good idea.
Be careful: the onchange should continue triggering modifications for prices etc.

elicoidal:
fixed by #132

PR #132, opened 2017-05-05
[FIX][business_requirement_deliverable] don't change deliverable description if previously filled

rubencabrera:
As described in #127

### DIFF
--- a/business_requirement_deliverable/models/business.py
+++ b/business_requirement_deliverable/models/business.py
@@ -233,7 +233,8 @@ class BusinessRequirementDeliverable(models.Model):
             )
             sale_price_unit = product.price
 
-        self.name = description
+        if not self.name:
+            self.name = description
         self.uom_id = uom_id
         self.sale_price_unit = sale_price_unit
 
--- a/business_requirement_deliverable/models/business.py
+++ b/business_requirement_deliverable/models/business.py
@@ -78,7 +78,8 @@ class BusinessRequirementResource(models.Model):
             uom_id = product.uom_id.id
         if product.description_sale:
             description += '\n' + product.description_sale
-        self.name = description
+        if not self.name:
+            self.name = description
         self.uom_id = uom_id
 
     @api.onchange('resource_type')
--- a/business_requirement_deliverable/tests/test_br.py
+++ b/business_requirement_deliverable/tests/test_br.py
@@ -211,7 +211,7 @@ class BusinessRequirementTestCase(common.TransactionCase):
         })
         self.br.write({'partner_id': self.partner.id})
         for line in self.br.deliverable_lines:
-            price_list = line._get_pricelist()
+            line.write({'product_id': self.productA.id, 'name': ''})
             self.assertEqual(
                 price_list.id,
                 self.partner.property_product_pricelist.id)
--- a/business_requirement_deliverable/tests/test_br.py
+++ b/business_requirement_deliverable/tests/test_br.py
@@ -282,7 +282,6 @@ class BusinessRequirementTestCase(common.TransactionCase):
                 sale_price_unit = product.price
 
             line.product_id_change()
-            self.assertEqual(line.name, description)
             self.assertEqual(line.uom_id.id, self.productA.uom_id.id)
             self.assertEqual(line.sale_price_unit, sale_price_unit)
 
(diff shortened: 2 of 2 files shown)


elicoidal:
@rubencabrera 
The magic of Travis! Can you please fix the test please?
Once done I will merge your contribution
Thanks a lot!

serpentcs-dev1:
@rubencabrera Then the same logic should be also applied [here,](https://github.com/OCA/business-requirement/blob/8.0/business_requirement_deliverable/models/business.py#L81) right?

serpentcs-dev1:
dlv new
But actually IMHO this code will work only once, onchange on description will not work if it is already set so then even if you change the product again the description would remain same.

edit : So travis is also failing for this reason.

martintorresvm:
> edit : So travis is also failing for this reason.

@darshan-serpent IMO is not the reason

Here we have the error on Travis and is true.
```
2017-05-06 10:22:52,006 5599 INFO openerp_test openerp.modules.loading: loading business_requirement_deliverable_cost/views/business.xml

2017-05-06 10:22:52,082 5599 
ERROR openerp_test openerp.addons.base.ir.ir_ui_view: 
Element '<xpath expr="//field[@name='unit_price']">' 
cannot be located in parent view

Error context:

View `business.requirement.deliverable.pivot`

[view_id: 861, xml_id: n/a, model: business.requirement.deliverable, parent_id: 853]
```
And here is the view inherited
```
        <!-- Pivot view Requirement Resources Line-->
        <record id="view_business_requirement_resource_pivot" model="ir.ui.view">
             <field name="name">business.requirement.resource.pivot</field>
             <field name="model">business.requirement.resource</field>
                 <graph string="Requirement Resources" type="pivot" stacked="True">
                     <field name="business_requirement_id" type="row"/>
                     <field name="business_requirement_deliverable_id" type="col"/>
                     <field name="qty" type="measure"/>
                 </graph>
             </field>
       </record>
```
I'll approve the changes.

martintorresvm (approved):

elicoidal:
@darshan-serpent 
> But actually IMHO this code will work only once, onchange on description will not work if it is already set so then even if you change the product again the description would remain same.

That is the intended behavior: having the description kept even if we change the product in the deliverable.
If the user wants to get the description of the product, he/she will have to empty first the description field.

> Then the same logic should be also applied here, right?

:+1: well spotted!

rubencabrera:
Sorry, been AFK for a few days. 
It is very annoying to change just one line of code and have to fix someone else's code, on a totally unrelated part, to make travis pass.
Any ideas if this error from the view can be fixed with a rebase or is it just a mistake that popped out now?

serpentcs-dev1:
@rubencabrera I guess you need to rebase first, though the travis would fail from [here](https://github.com/OCA/business-requirement/blob/8.0/business_requirement_deliverable/tests/test_br.py#L247) as you know because the onchange will not populate the name field until it is empty.

serpentcs-dev1:
We fixed the travis issue from [here](https://github.com/OCA/business-requirement/blob/8.0/business_requirement_deliverable/tests/test_br.py#L211) by changing it to : 
`            line.write({'product_id': self.productA.id, 'name': ''})
`

martintorresvm:
> Sorry, been AFK for a few days.
It is very annoying to change just one line of code and have to fix someone else's code, on a totally unrelated part, to make travis pass.
Any ideas if this error from the view can be fixed with a rebase or is it just a mistake that popped out now?

@rubencabrera Indeed is it.

Could you rebase this PR and check if Travis passed this time?

rubencabrera:
Rebased and there's nothing to update. Tried @darshan-serpent's suggestion for travis fix.

serpentcs-dev1:
@rubencabrera Please check the above link that I have shared the line you have made the change is not in the expected method.

rubencabrera:
Thank you @darshan-serpent , I guess I also did something wrong with the rebase, because when rebasing from 8.0 it said nothing to update, but my line 211 was the one I changed on the previous wrong commit. Sorry for the inconvenience.

elicoidal:
Is this ready to merge?
@rubencabrera

rubencabrera:
I screwed the rebase, let me tidy this mess a little bit

serpentcs-dev1:
@rubencabrera  Yeah, I think your branch has diverged, maybe you need to reset to OCA and then commit it again :)

rubencabrera:
Rebased done, sorry for the delay.

elicoidal:
Thanks Ruben, I fixed as well line 81.
Merging

elicoidal:
Travis is unrelated. Fast tracking this one

PR #135, opened 2017-05-12
[FIX] business_requirement_deliverable

martintorresvm:
- Fix double factor conversion when UoM
- Fix Total revenue should be recomputed when sales price on DL changes


martintorresvm:
@elicoidal Could be merge this as well?

elicoidal (approved):

elicoidal:
Fasttracking

PR #142, opened 2017-05-19
[FIX] business_requirement_deliverable_cost

martintorresvm:
- Fix unit_price Issue computed fields are related and cause error


martintorresvm:
@elicoidal @seb-elico This fix the error caused by with `unit_price` on BR with BRD.Cost module for those who doesn't belong to Cost group

elicoidal (approved):
Fast tracking

Issue #146, opened 2017-05-25
Field Linked project should be readonly


elicoidal:

rubencabrera:
Isn't it already readonly? It's only editable on draft state, both on business_requirement and business_requirement_deliverable models. It seems to work as defined on my test db, can you give more detail to reproduce?

elicoidal:
@rubencabrera even in draft state it should be read only.
This field is be automatically filed in when we generate the project.

I don't think there is a need to let the user input manually that information: do you have use case that would support to keep it not read only in draft (or more stages)?

rubencabrera:
I don't see the need for it to be editable, I also prefer the automatic link when generating project.

elicoidal:
Thanks for your feedback. We will proceed then

rubencabrera:
Should it be readonly too for the deliverable?

elicoidal:
yes: although it is not visible in form view

rubencabrera:
Fine, I assumed that, you can test the PR to see if it fits.

elicoidal:
Fixed by #148

PR #139, opened 2017-05-16
[WIP] business_requirement_deliverable_report

martintorresvm:
- Add some of the features of V9 branch within reports

PR #148, opened 2017-05-25
Readonly always for linked_project

rubencabrera:
As discussed in #146

### DIFF
--- a/business_requirement_deliverable/models/business.py
+++ b/business_requirement_deliverable/models/business.py
@@ -78,6 +78,8 @@ class BusinessRequirementResource(models.Model):
             uom_id = product.uom_id.id
         if product.description_sale:
             description += '\n' + product.description_sale
+        if not self.name:
+            self.name = description
         self.uom_id = uom_id
 
     @api.onchange('resource_type')
--- a/business_requirement_deliverable/models/business.py
+++ b/business_requirement_deliverable/models/business.py
@@ -78,8 +78,6 @@ class BusinessRequirementResource(models.Model):
             uom_id = product.uom_id.id
         if product.description_sale:
             description += '\n' + product.description_sale
-        if not self.name:
-            self.name = description
         self.uom_id = uom_id
 
     @api.onchange('resource_type')
--- a/business_requirement_deliverable_project/models/business.py
+++ b/business_requirement_deliverable_project/models/business.py
@@ -20,7 +20,6 @@ class BusinessRequirement(models.Model):
         comodel_name='project.project',
         groups='project.group_project_user',
         readonly=True,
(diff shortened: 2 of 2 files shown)


elicoidal (approved):

elicoidal:
@rubencabrera Can you please check travis?

rubencabrera:
Sure, I'm onto it, seems to be something to do with the product_id onchange, seems related to the recently merged PR, but it was solved there. Let me check.

rubencabrera:
The test was right, that product_id onchange should still change the description (name field). The mentioned pr was for the deliverable, not the resource. So this last commit also fixes an error of PR #132

rubencabrera:
Ok, the travis fails because a test of business_requirement_deliverable_resource, I was going to change it but I'm not sure which is the expected behaviour.
The test calls product_id_change and asserts that the resource name is equal to the product name, but the function doesn't do that. Should we change it? To do exactly what?

elicoidal:
Not sure either...
@victormartinelicocorp could you have a quick look?

business_requirement_deliverable/models/business.py
```diff
@@ -78,8 +78,6 @@ def product_id_change(self):
             uom_id = product.uom_id.id
         if product.description_sale:
             description += '\n' + product.description_sale
-        if not self.name:
```
elicoidal: @rubencabrera 
After reconsidering: why should we have this change in this PR?
I think you need to separate this modification that was done in another PR (may rebase)?
@victormartinelicocorp @darshan-serpent

elicoidal (changes requested):

business_requirement_deliverable/models/business.py
serpentcs-dev1: @rubencabrera As you know the code for onchange on the resource is already merged in #132 
and the test cases to support it were already added in Open PR #139 check this commit
https://github.com/OCA/business-requirement/pull/139/commits/2a02aedc74743772bff01b9e178b4408f01eae1f

So the travis is already fixed in the above mentioned PR.

So IMO you just need to revert your last commit and that will do all.

cc @elicoidal @victormartinelicocorp

rubencabrera: That's what I thought, just wanted to get rid of the annoying travis failing for outside code. I'll revert and commit ASAP. Thanks again to all of you for your input.

darshan-serpent:
+1

business_requirement_deliverable/models/business.py
elicoidal: @rubencabrera any news?

elicoidal (approved):

elicoidal:
Thanks!

Issue #157, opened 2017-05-26
In RL tree view: UoM should be after the Qty. Sequence should be the first column


elicoidal:
@darshan-serpent

elicoidal:
#159

PR #159, opened 2017-05-26
[FIX] 8.0-Issue-157-RL-Treeview

serpentcs-dev1:
Fixes Issue : https://github.com/OCA/business-requirement/issues/157

In RL tree view: Set the UoM after the Qty and Sequence on the first column


elicoidal:
Travis fails unrelated to this PR (to be fixed in separate PR)
Fast tracking as trivial

elicoidal (approved):

Issue #150, opened 2017-05-26
The BR form view does not have cost control tab anymore


elicoidal:
This tab existed in previous version with following information:
Total Revenue (from deliverable)
Total cost for task (from resources)
Total cost for Procurement (from resources)
Total Gross Profit

elicoidal:
@darshan-serpent

PR #164, opened 2017-05-30
[FIX] The BR form view does not have cost control tab anymore

serpentcs-dev1:
Fixes issue : - #150 

Please review

@elicoidal


elicoidal (approved):
Trivial. Fast tracking

Issue #144, opened 2017-05-25
In stages missing the stage "drop"


elicoidal:
We need 2 stages like in previous version:
* Cancel: the demand is irrelevant/erroneous
* Drop: decision is made not to pursue the BR

Currently only one (Cancel)

elicoidal:
Actually the stage exists and can be used in kanban view but does not appear in the form view

elicoidal:
@darshan-serpent

serpentcs-dev1:
Okay let me check

PR #165, opened 2017-05-30
[FIX]144 In stages missing the stage drop

serpentcs-dev1:
Fixes :- #144 (In stages missing the stage "drop" )

Please review it

@elicoidal


elicoidal (approved):
Trivial. Fast tracking

Issue #160, opened 2017-05-27
In BR move the 2 fields Requested by and responsible just below the name


elicoidal:
Besides, change the order: Responsible first then Requested by

@darshan-serpent

PR #166, opened 2017-05-30
[IMP]Improved BR-fields position

serpentcs-dev1:
#160 

Moved the 2 fields Requested by and responsible below BR name.

@elicoidal


elicoidal (approved):
Trivial: fast tracking

Issue #161, opened 2017-05-27
In BR form view, it should be possible to change the kanban state without editing the form


elicoidal:
@darshan-serpent

elicoidal:
@darshan-serpent

serpentcs-dev1:
Yes, I am working on it. Actually not able to find the reason maybe it needs a JS tweak.

elicoidal:
It should be similar in behavior to the project or task widget no?

serpentcs-dev1:
Yeah right tried to apply the same behavior but does not work in this case, strange!

No problem I'll double check it and fix it.

serpentcs-dev1:
@elicoidal I was testing I found that the kanban_state stays non-editable if it is the inherited view.

But if I add the field `kanban_state` in business_requirement module it works smooth as silk.

elicoidal:
@darshan-serpent thanks for the feedback!
Strange but if so then can you transfer the field accordingly?

serpentcs-dev1:
Yes I would do that, just let try this and let's see how it looks on runbot, till then I'll try to also find the reason from the root.

rubencabrera:
Someone please close so we don't keep checking it. Thanks @darshan-serpent for the PR!

elicoidal:
Thanks for the follow up!

PR #169, opened 2017-06-06
[8.0] Issue-161 Kanban state editable

serpentcs-dev1:
This PR resolves the Kanban State issue #161 to make it editable from the form view of BR.

This is the simple workaround which moves the `kanban_state` field from the inherited br_project form view to the BR module main view (as for some reason it was not able to edit the field in inherited form view)

P.S: Need to find a more logical way

### DIFF
--- a/business_requirement/models/business.py
+++ b/business_requirement/models/business.py
@@ -197,6 +197,13 @@ class BusinessRequirement(models.Model):
     to_be_reviewed = fields.Boolean(
         string='To be Reviewed'
     )
+    kanban_state = fields.Selection([('normal', 'In Progress'),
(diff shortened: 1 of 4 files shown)


elicoidal (commented):
Not ideal but I understand the workaround
@pedrobaeza @dreispt this is to answer #161 

Anything we are missing?

dreispt (approved):
LGTM

dreispt:
AFAICS it makes sense for all Stage/Sate/Kanvban State fields to be defined in `business_requirement` instead of  `business_requirement_deliverable_project`.

elicoidal:
@dreispt Thanks for the follow up.
Merging

PR #172, opened 2017-06-08
[FIX] business_requirement_deliverable and business_requirement_deliverable_cost

martintorresvm:
[FIX] business_requirement_deliverable_cost
- Actual ACL configuration makes imposible for a BR user
create,modify,duplicate properly a BR with the modules:
  - business_requirement_deliverable
  - business_requirement_deliverable_cost
- Need to move the ACL restrictions on View level.

### DIFF
--- a/business_requirement_deliverable/models/business.py
+++ b/business_requirement_deliverable/models/business.py
@@ -283,9 +283,7 @@ class BusinessRequirement(models.Model):
     total_revenue = fields.Float(
         compute='_compute_deliverable_total',
         string='Total Revenue',
-        store=True,
-        groups='business_requirement_deliverable.'
-        'group_business_requirement_estimation'
+        store=True
     )
     currency_id = fields.Many2one(
         comodel_name='res.currency',
--- a/business_requirement_deliverable/views/business_view.xml
+++ b/business_requirement_deliverable/views/business_view.xml
@@ -239,7 +239,9 @@
             <field name="arch" type="xml">
                 <xpath expr='//tree/field[@name="change_request"]' position='after'>
                     <field name="currency_id"/>
-                    <field name="total_revenue" sum="Total revenue"/>
+                    <field name="total_revenue"
+                           sum="Total revenue"
+                           groups="business_requirement_deliverable.group_business_requirement_estimation"/>
                 </xpath>
             </field>
         </record>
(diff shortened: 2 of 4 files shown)


martintorresvm:
@elicoidal @seb-elico Could have a look?

dreispt (approved):

rubencabrera (approved):
:+1:

elicoidal (approved):

elicoidal:
@seb-elico @dreispt @pedrobaeza 

I personally prefer the control at model level (to avoid export or ERPPeek leaks) but I do not see how to achieve reasonable usability.

Shall we override the copy function and use SUDO instead?

elicoidal (commented):
@seb-elico @dreispt @pedrobaeza

I personally prefer the control at model level (to avoid export or ERPPeek leaks) but I do not see how to achieve reasonable usability.

Shall we override the copy function and use SUDO instead?

martintorresvm:
@elicoidal To override functions is not only for the `copy` function also need to be  done for `create` and `write` and take care to rewrite the `create_uid` and `write_uid` with the original user after `sudo()`. 

About security:

IMHO a user just for have access to a production server should be considered a the main vulnerability even if the user doesn't have the rights on Odoo to read these fields by using erppeek scripts. 

Just for example if could have access to the postgres DB could hack very easy `admin` password and whatever happens is depends on the intention.

To avoid the export:

We could set of a list of fields that if exist on the `fields_to_export` from the `export_data` function then need to check if belong to these groups.

elicoidal:
@victormartinelicocorp I understand the limit of SUDO and yes we can tweak the export data function but for ERPPeek, no need to have admin or Postgres access and there is still a risk.

dreispt:
AFAICS the root of the problem is that you want is to limit write access on specific fields.
The problem with `groups` on a field is that it makes the field totally inaccessible to other users, so its not fit for that job.

While the current solution is a good workaround, it does not ensure actual security on those fields, as Eric pointed out.
The proper solution is to control the write on those fields at the ORM level, in the `create`/`write` methods.
That could be done on top of this PR.

elicoidal:
OK Merging. 
Another PR will be prepare to cap the users rights on read/write and delete functions.

Issue #145, opened 2017-05-25
Cannot change assignee in Resource lines in Stakeholder approval stage


elicoidal:

elicoidal:
@darshan-serpent

serpentcs-dev1:
I guess I missed this, will continue to work on this.

serpentcs-dev1:
@elicoidal Here in RL we need state field from BR to make them readonly / editable. So I guess, we should create a related field of **state** of BR and use it in XML in attrs.

Let me know this is a right way to do this or there is any other way to achieve this functionality.

elicoidal:
It is reasonable to have in RL  information related/stored (depending) from the BR (customer, state, pricelist etc.)
So for me :+1:

PR #173, opened 2017-06-08
[IMP]-Improved assignee editable issue.

serpentcs-dev1:
#145 This PR resolves the Assignee editable issue.

**[WIP] Improving code**

Changes made : 

- Added a related field to apply the readonly attribute on the fields 
- Removed the bogus/redundant code which was having two views for the BR deliverable and resource

### DIFF
--- a/business_requirement_deliverable/models/business.py
+++ b/business_requirement_deliverable/models/business.py
@@ -66,6 +66,8 @@ class BusinessRequirementResource(models.Model):
         string='Business Requirement',
         store=True
     )
+    state = fields.Selection(related='business_requirement_id.state',
+                             string='State', store=True)
 
     @api.multi
     @api.onchange('product_id')
@@ -166,6 +168,8 @@ class BusinessRequirementDeliverable(models.Model):
         string='Business Requirement',
         store=True
     )
+    state = fields.Selection(related='business_requirement_id.state',
+                             string='State', store=True)
 
     @api.multi
     @api.onchange('business_requirement_id')
@@ -297,8 +301,6 @@ class BusinessRequirement(models.Model):
                                     compute='_compute_dl_total_revenue')
     dl_count = fields.Integer('DL Count', compute='_compute_dl_count')
     rl_count = fields.Integer('RL Count', compute='_compute_rl_count')
-    dl_count_noedit = fields.Integer('DL Count', compute='_compute_dl_count')
-    rl_count_noedit = fields.Integer('RL Count', compute='_compute_rl_count')
 
     @api.multi
     def _compute_dl_total_revenue(self):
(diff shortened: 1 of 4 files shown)


serpentcs-dev1:
@elicoidal Could you please review?

business_requirement_deliverable/views/business_view.xml
```diff
@@ -321,34 +201,44 @@
...
                         nolabel="1">
-                        <tree editable="1">
-                            <field name="resource_type"/>
-                            <field name="product_id"/>
-                            <field name="name"/>
-                            <field name="user_id" attrs="{'readonly':[('resource_type','=','procurement')]}"/>
-                            <field name="qty"/>
-                            <field name="uom_id"/>
-                            <field name="business_requirement_id" invisible="1"/>
+                        <tree editable="1" >
```
dreispt: It's best to leave the Tree view defined it it's own record, rather than defining it inside a field - it makes it easier to extend.

elicoidal:
@darshan-serpent How to test your modifications?

serpentcs-dev1:
I guess it is the just like the way we were doing before. 

- To create a BR having deliverable and resource lines, which would be editable in 'Draft' and 'Confirm' State
- Then to change the state to 'Stakeholder Approval' which will then only allow to change the `'Assigned To'` field which is on resource lines.

elicoidal (approved):
Tests LGTM.
Can you attend @dreispt concerns?

business_requirement_deliverable/views/business_view.xml
serpentcs-dev1: @dreispt  Yes right, but here we want to display only specific fields having attributes that only remains to this view :)

@elicoidal

elicoidal: Not sure that I can help here except that I can lose the requirement to make it more flexible later

serpentcs-dev1:
@dreispt 
Regarding your comment, I think but here we want to display some specific fields having attributes that only remains to this view so we have added an inner tree view.

What do you suggest? Need your review!!

elicoidal:
@dreispt @rubencabrera 
Do you have any more feedback? Otherwise I would like to move forward with the PR

serpentcs-dev1:
@elicoidal What do you think about this PR?? 

This PR resolves several other issues like https://github.com/OCA/business-requirement/issues/191

elicoidal:
@victormartinelicocorp @pedrobaeza Could we move forward as @dreispt seems out of office?

business_requirement_deliverable/models/business.py
```diff
@@ -166,6 +168,8 @@ class BusinessRequirementDeliverable(models.Model):
         string='Business Requirement',
         store=True
     )
+    state = fields.Selection(related='business_requirement_id.state',
+                             string='State', store=True)
```
pedrobaeza: make it readonly=True or you can have problems

pedrobaeza (changes requested):

business_requirement_deliverable/models/business.py
serpentcs-dev1: Yes, right!

serpentcs-dev1:
@pedrobaeza Your request is attended, could you please review it again?!

pedrobaeza (approved):

elicoidal:
@dreispt This points are really blocking for us now so I propose that we move forward.
@darshan-serpent can you add a comment in the code with Daniel's concerns and we merge?

serpentcs-dev1:
@elicoidal comment added.

elicoidal:
Thanks @darshan-serpent

Issue #170, opened 2017-06-08
[business_requirement_deliverable_report] BR print report is not printing all BR selected on Tree view


martintorresvm:
On the module: business_requirement_deliverable_report

Have a bug on print reports for more than one BR selected.

To reproduce the Issue just need to:

1. click on BR tree view and select more than one.
2. Then on click onPrint  menu button. 
3. Select the one report (eg:  Bussiness Requirment Resources Report)
4. Will generate a report just with one BR. 

Expected print all BR selected.

martintorresvm:
@elicoidal

elicoidal:
@darshan-serpent

serpentcs-dev1:
@victormartinelicocorp I tested the steps you mentioned I guess this is because the [Attachement is set to True.](https://github.com/OCA/business-requirement/blob/8.0/business_requirement_deliverable_report/report/report.xml#L12)
 Try deleting it once and then you would be able to print the new one. It worked for us.
cc @elicoidal 


![screenshot from 2017-06-14 15-46-19](https://user-images.githubusercontent.com/20109026/27127534-08aa71b0-5119-11e7-8f86-9e584b89dd6b.png)

martintorresvm:
@darshan-serpent Thanks. Then we'll see if need to set to false because for user is annoying that.

cc @seb-elico @elicoidal

seb-elico:
@Vic agreed let's not set it to `True` if that fixes the bug!

elicoidal:
@darshan-serpent can you create a PR for that?

serpentcs-dev1:
@elicoidal Okay, I'll create a new PR, which removes the attachment facility!

elicoidal:
@darshan-serpent Can you link the related PR?

PR #182, opened 2017-06-23
[FIX] attachment false

serpentcs-dev1:
@elicoidal  Removes Attachment functionality, refers to issue #170 

cc @victormartinelicocorp


elicoidal (approved):
Fast track

Issue #184, opened 2017-06-27
[RFC]Reference field improvements


elicoidal:
Today the Reference field usage is not clear and could be misinterpreted.
* Originally usage was to input a WBS or reference from the customer
* Name get removes the BR number and replace it by the Reference
Issue:
* A unique BR numbering is a mandatory in day to day work for consistency
* It is not clear the usage between Reference, WBS, Source (as additional link)

Proposal for improvement:
* Rename Reference field into WBS
* Change the name_get of the BR into [WBS][BRNUMBER] BRName (if WBS is not empty)
* Create a new field Source (searchable, name=origin to follow Odoo conventions) where the users can input any free text. To be added after Customer

elicoidal:
@rubencabrera @dreispt @jbeficent @victormartinelicocorp @seb-elico

PR #208, opened 2017-07-03
[8.0]-[BR1030]-[ADD/IMP][Reference field improvements]

serpentcs-dev1:
Issue:- #184 

- Improved string Reference to "WBS"

- Improved name_get method

- Added origin field with text type

### DIFF
--- a/business_requirement/models/business.py
+++ b/business_requirement/models/business.py
@@ -40,7 +40,7 @@ class BusinessRequirement(models.Model):
         states={'draft': [('readonly', False)]}
     )
     ref = fields.Char(
-        'Reference',
+        'WBS',
         required=False,
         readonly=True,
         copy=False,
@@ -204,6 +204,7 @@ class BusinessRequirement(models.Model):
                                     track_visibility='onchange',
                                     required=False,
                                     copy=False, default='normal')
+    origin = fields.Text(string='Source')
 
     @api.multi
     @api.onchange('project_id')
@@ -269,7 +270,8 @@ class BusinessRequirement(models.Model):
         result = []
         for br in self:
             if br.ref:
-                formatted_name = u'[{}] {}'.format(br.ref, br.description)
+                formatted_name = u'[{}][{}] {}'.format(br.ref, br.name,
+                                                       br.name)
             else:
                 formatted_name = u'[{}] {}'.format(br.name, br.description)
             result.append((br.id, formatted_name))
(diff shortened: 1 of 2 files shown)


business_requirement/models/business.py
```diff
@@ -40,7 +40,7 @@ def _get_default_company(self):
         states={'draft': [('readonly', False)]}
     )
     ref = fields.Char(
-        'Reference',
+        'WBS',
         required=False,
         readonly=True,
```
elicoidal, line 45: Any reason why r/o?

elicoidal (approved):
LGTM 
@victormartinelicocorp @seb-elico

business_requirement/models/business.py
```diff
@@ -269,7 +270,8 @@ def name_get(self):
         result = []
         for br in self:
             if br.ref:
-                formatted_name = u'[{}] {}'.format(br.ref, br.description)
+                formatted_name = u'[{}][{}] {}'.format(br.ref, br.name,
```
martintorresvm, line 273: @darshan-serpent What is the point of have twice the br.name? Perhaps you forgot add other field?

martintorresvm (changes requested):

business_requirement/models/business.py
serpentcs-dev1: Hmm.. I haven't changed it was already there.

But yes, I think read-only should not be there, should I remove?

business_requirement/models/business.py
serpentcs-dev1: Yes you are right. I think It should be like, right? 

`formatted_name = u'[{}] {}'.format(br.ref, br.name, br.description)`

Will do the needful, thanks!

business_requirement/models/business.py
elicoidal: I am not sure as there are some ACL based on status. I dont know if it has an impact so dont modify it here.

business_requirement/models/business.py
elicoidal: not sure here but if the requirement is fulfilled, this is fine by me.

I want: [WBS][name] Description or [name] Description (if WBS empty)

martintorresvm: :+1:

serpentcs-dev1:
@elicoidal @victormartinelicocorp Tested on runbot, works as expected.

The test is made with 2 BRs, one with WBS and other without it.

without_wbs
with_wbs

elicoidal (approved):
Perfect!

elicoidal:
@victormartinelicocorp please review again

martintorresvm (approved):

Issue #189, opened 2017-06-29
Add a button to copy the total revenue in Deliverable Lines


elicoidal:
Currently in the DL form view, the list of RL are providing an automatic calculation based on the resources and sales prices at RL level.
Total revenue at resource line level is displayed as a total.

For business reasons, this total is not updated automatically to the Deliverable lines: there might be multiple reasons not to apply the automatic calculation.

To ease though the data entry we suggest to add a button to copy on demand the total revenue calculated in the RL into the DL revenue.

serpentcs-dev1:
@elicoidal 

So you need a button in the DL which will just sum the total revenue from all the RL and will propagate it to the DL's total revenue??

 Is this what you mean?

elicoidal:
as simple as that yes. Only for the given deliverable.
@darshan-serpent

serpentcs-dev1:
Still, one little thing to clear.

There's already a field in DL called Total Revenue, which right now is computed as `sale_price * qty`

So now... should we append the calculation of RL total revenue to the DL revenue field or totally should replace??

![deliverable lines odoo](https://user-images.githubusercontent.com/20109026/27818342-def02c08-60b2-11e7-8f76-5cd79571050f.png)

elicoidal:
@darshan-serpent https://github.com/OCA/business-requirement/pull/211#issuecomment-312795556
I have clarified my initial requirement: it should update the Sales Price field.
Let me know if still unclear

PR #211, opened 2017-07-04
[FIX] [8.0-i189-BR001030-T22334] Add button to copy the total revenue in Deliverable Lines

sudhir-serpentcs:
@elicoidal @seb-elico @victormartinelicocorp 
#189 
Added a button then copy total revenue in Deliverable Lines.

### DIFF
--- a/business_requirement_deliverable_cost/models/business.py
+++ b/business_requirement_deliverable_cost/models/business.py
@@ -159,11 +159,8 @@ class BusinessRequirementDeliverable(models.Model):
 
     @api.multi
     def action_button_update_total_revenue(self):
-        total_revenue = 0.0
-        if self.price_total != total_revenue:
-            for rl in self.resource_ids:
-                total_revenue += rl.sale_price_total
-            self.price_total = total_revenue
+        self.sale_price_unit = sum(self.resource_ids.
+                                   mapped('sale_price_total'))
 
 
 class BusinessRequirement(models.Model):
--- a/business_requirement_deliverable_cost/models/business.py
+++ b/business_requirement_deliverable_cost/models/business.py
@@ -157,6 +157,14 @@ class BusinessRequirementDeliverable(models.Model):
                     resource._set_sales_price()
                     resource._set_cost_price()
 
+    @api.multi
+    def action_button_update_total_revenue(self):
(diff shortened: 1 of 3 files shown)


business_requirement_deliverable_cost/models/business.py
```diff
@@ -157,6 +157,14 @@ def action_button_update_estimation(self):
                     resource._set_sales_price()
                     resource._set_cost_price()
 
+    @api.multi
+    def action_button_update_total_revenue(self):
+        total_revenue = 0.0
+        if self.price_total != total_revenue:
+            for rl in self.resource_ids:
```
elicoidal: You can probably find a more pythonic way to achieve this but I am fine with it

business_requirement_deliverable_cost/models/business.py
```diff
@@ -157,6 +157,14 @@ def action_button_update_estimation(self):
                     resource._set_sales_price()
                     resource._set_cost_price()
 
+    @api.multi
+    def action_button_update_total_revenue(self):
+        total_revenue = 0.0
+        if self.price_total != total_revenue:
```
elicoidal: This check does not make so much sense to me as total_revenue is always 0, does it?

elicoidal:
The button should be located just before or after the other button at the top.
Currently not working but because I think I was not clear enough: it should copy the total revenue in the field "Sales Price" @sudhir-serpentcs

business_requirement_deliverable_cost/models/business.py
martintorresvm: @sudhir-serpentcs No need to check condition here. If user click on the button is because want to update the value.

business_requirement_deliverable_cost/models/business.py
martintorresvm: @sudhir-serpentcs Think you could use https://docs.python.org/2/library/functions.html#map function.

martintorresvm (changes requested):

sudhir-serpentcs:
Updated the sale price field using mapped function.

martintorresvm (approved):

elicoidal (approved):
Functional tests OK

PR #224, opened 2017-07-26
[MIG]-Migrated Business Requirement Deliverable from 8.0 to 10.0.

sudhir-serpentcs:
Issue #48

- This PR is depend on #223.
- Migrated Business Requirement Deliverable from v8 to v10.

PR #223, opened 2017-07-26
[MIG]Migrated Business requirement module from 8.0 to 10.0

YogeshMahera-SerpentCS:
#48 

- Migrated Business Requirement from v8 to v10

Issue #48, opened 2016-10-06
Migration to version 10.0 [FINISHED]


pedrobaeza:
# Todo

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-10.0
# Modules to migrate
When you pick a module, input your github id and the PR number after the name of the module:
- [x] business_requirement #223 
- [x] business_requirement_deliverable #224 
- [X] business_requirement_deliverable_categ (DEPRECATED)
- [x] business_requirement_deliverable_cost #225 
- [x] business_requirement_deliverable_crm #229 
- [x] business_requirement_deliverable_default (DEPRECATED)
- [x] business_requirement_deliverable_project #231 
- [X] business_requirement_deliverable_project_categ (DEPRECATED)
- [x] business_requirement_deliverable_project_task_categ #233 
- [x] business_requirement_deliverable_report (TO BE MOVED TO EACH SPECIFIC MODULE)
- [x] business_requirement_deliverable_resource_template #236 
- [x] business_requirement_deliverable_resource_template_categ #237
- [x] business_requirement_earned_value #234 (DEPRECATED)
- [x] business_requirement_etherpad #235

elicoidal:
All good!. Migration Finished

pedrobaeza:
Congratulations!

PR #228, opened 2017-07-31
[10.0]-[WIP]-Migrated Business Requirement Deliverable Categ from 8.0 to 10.0.

sudhir-serpentcs:
Issue #48 

- This PR depends on #223 and #224 
- Migrated Business Requirement Deliverable from v8 to v10.
- This module is dependent on "**[project_categ](https://github.com/oca/project)**" module but it's not migrated in new API

### DIFF
--- a/business_requirement_deliverable_categ/README.rst
+++ b/business_requirement_deliverable_categ/README.rst
@@ -9,7 +9,7 @@ Business Requirement Resources Task Categories
 Introduction
 ============
 
-This module is part of a set of modules (`Business Requirements <https://github.com/OCA/business-requirement/blob/8.0/README.md>`_).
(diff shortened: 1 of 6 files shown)


elicoidal:
@victormartinelicocorp do we need to migrate this one? Isnot it the one which is deprecated?

martintorresvm:
> @victormartinelicocorp do we need to migrate this one? Isnot it the one which is deprecated?

@elicoidal You are right, deprecated one.

elicoidal:
@sudhir sorry about that :smile:

sudhir-serpentcs:
@elicoidal @victormartinelicocorp So we don't need to migrate "business_requirement_deliverable_categ" and "project_categ" module anymore in v10?

elicoidal:
neither do you have to upgrade 'business_requirement_deliverable_project_categ: I have update the list according in #48

PR #232, opened 2017-08-02
[MIG]Migrated Business Requirement Deliverable Report from 8.0 to 10.0.

sudhir-serpentcs:
Issue #48 

- This PR is depend on #224 .
- Migrated Business Requirement Deliverable from v8 to v10.

### DIFF
--- a/business_requirement_deliverable_report/README.rst
+++ b/business_requirement_deliverable_report/README.rst
@@ -9,7 +9,7 @@ Business Requirement Document Printout
 Introduction
 ============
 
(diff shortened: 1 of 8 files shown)


elicoidal (approved):
LGTM. Pending testing in Runbot

elicoidal (changes requested):
Actually, this module should be deprecated. The reports should be transfered to the corresponding modules: brd (1) and brd_cost (2)

sudhir-serpentcs:
> elicoidal requested changes 8 hours ago
Actually, this module should be deprecated. The reports should be transfered to the corresponding modules: brd (1) and brd_cost (2)

@elicoidal  Sure, I will do it at the end when other modules are migrated and then will add the reports to respected modules.

PR #243, opened 2017-08-19
Kanban view move from BRP to BR

sudhir-serpentcs:
#224 
- Kanban view move from BRP to BR module


elicoidal:
Thanks @sudhir-serpentcs please find my comments:
- [ ] when creating a new BR, focus is on Use case, not on name.
- [x] no drag and drop in kanban.

YogeshMahera-SerpentCS:
@elicoidal could you please review.

YogeshMahera-SerpentCS:
@victormartinelicocorp 
> when creating a new BR, focus is on Use case, not on name.

For that, I have added default_focus=1, While open BR form it first redirects to the description, but when I again open BR form or Refresh the same page redirect to use case field.

https://github.com/sudhir-serpentcs/business-requirement/blob/10.0-PR-REF-224/business_requirement/views/business_view.xml#L39

https://github.com/sudhir-serpentcs/business-requirement/blob/10.0-PR-REF-224/business_requirement/views/business_view.xml#L64

can you please check it give us feedback.
cc @elicoidal @sudhir-serpentcs

elicoidal:
FYI it was fixed in v8 after we had the same issue

YogeshMahera-SerpentCS:
I checked all the commit history of the  version 8 and use various other possible ways using 'tabindex' or 'default_focus'.
but this doesn't seems to work

elicoidal:
@pedrobaeza @dreispt @lasley we are looking for some advice here.
Issue is: when creating a new business requirement, the focus get set on the Use case field instead of the name field.
We cannot pinpoint any reason: any idea where to search?

elicoidal (approved):
Let's move ahead with this one. We will fix the focus later on.

elicoidal:
@victormartinelicocorp can you review?

lasley:
Can you add the focus issue to ReadMe & add the tactics that have been attempted. I tried all the standard ones too, which I'm sure you also tried, to no avail

elicoidal:
@lasley strange enough is that we had the same issue in v8 and it got it solve along the road :grimacing: 
We searched the history but could not find the reason neither there.

elicoidal:
@YogeshMahera-SerpentCS can you attend following comment too: https://github.com/OCA/business-requirement/pull/243#issuecomment-327210508

YogeshMahera-SerpentCS:
@elicoidal Yes, Added Issue in README.

elicoidal:
@YogeshMahera-SerpentCS thanks
@lasley if you dont mind I would like to move forward.
@victormartinelicocorp Can you review this too?

martintorresvm (approved):

elicoidal:
Moving forward with this one
@YogeshMahera-SerpentCS can you rebase the other PR?

Issue #253, opened 2017-09-21
In BR the cost control calculation is incorrect


elicoidal:
Current behavior
summary of the DL/RL:
![screenshot from 2017-09-21 11-04-11](https://user-images.githubusercontent.com/7600613/30677166-b38542cc-9ebc-11e7-80d7-b790041ebf12.png)

In BR the cost control (company currency is CNY):
Currency | EUR
-- | --
Total Revenue | 178,540.65
Total procurement | 3,158.00
Total tasks | 34,960.00
Estimated Gross Profit | 140,422.65

Expected behavior:
- Total revenue is based on the total revenue in the DL
- A new field Total Revenue in company Currency to be added below with the proper currency conversion

elicoidal:
@YogeshMahera-SerpentCS

YogeshMahera-SerpentCS:
@elicoidal 
Fixed issue in v8 and v10 #225 and #255

elicoidal:
@YogeshMahera-SerpentCS this is milestones 10.0

PR #255, opened 2017-09-21
[8.0]-[IMP]-[FIX]-Fixed issues of 253

YogeshMahera-SerpentCS:
#253

### DIFF
--- a/business_requirement_deliverable_cost/models/business.py
+++ b/business_requirement_deliverable_cost/models/business.py
@@ -228,7 +228,7 @@ class BusinessRequirement(models.Model):
                         .filtered(lambda r: r.resource_type == 'task')
                         .mapped('price_total'))
                 resource_task_total_ci = self.env['res.currency']._compute(
-                    self.currency_id, self.env.user.company_id.currency_id,
+                    rec.currency_id, self.env.user.company_id.currency_id,
                     resource_task_total)
                 br.resource_task_total = resource_task_total_ci
 
--- a/business_requirement_deliverable_cost/models/business.py
+++ b/business_requirement_deliverable_cost/models/business.py
@@ -205,8 +205,8 @@ class BusinessRequirement(models.Model):
     def _compute_total_revenue_in_ci(self):
         for rec in self:
             total_revenue_ci = self.env['res.currency']._compute(
-                self.currency_id, self.env.user.company_id.currency_id,
-                self.total_revenue)
+                rec.currency_id, self.env.user.company_id.currency_id,
+                rec.total_revenue)
             rec.total_revenue_ci = total_revenue_ci
 
     @api.multi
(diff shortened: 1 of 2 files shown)


YogeshMahera-SerpentCS:
@elicoidal 
Can you please review it

elicoidal:
@YogeshMahera-SerpentCS it's aslmost perfect! 
* Just name to be changed from "Total Revenue in CI" to "Total Revenue (Cie Curr.)"
* "Total Revenue (Cie Curr.)" +amount+field should only be displayed if company currency is different from PL currency

Please apply the same methodology for the deliverable lines:
* Remove field currency (but keep the symbol)
* Add new field "Total Revenue (Cie Curr.)"
* Add the company currency symbol
* "Total Revenue (Cie Curr.)" +amount+field should only be displayed if company currency is different from PL currency

business_requirement_deliverable_cost/models/business.py
```diff
@@ -185,6 +185,29 @@ class BusinessRequirement(models.Model):
         'RL Total Cost',
         compute='_compute_rl_total_cost'
     )
+    total_revenue_ci = fields.Float(
+        compute='_compute_total_revenue_in_ci',
+        string='Total Revenue in CI',
+    )
+    ci_currency_id = fields.Many2one(
```
martintorresvm, line 192: `ci_` prefix would be means nothing as name for a field, better add what functional means on the `help` parameter

- [ ] Add `help` param with a short meaning
- [ ] Then name then should be `currency_id`
- [ ] Remove the readonly
- [ ] Should be a `default` instead of `compute` from the br.self.partner_id.property_product_estimation_pricelist.currency_id
but later the user what change then could.

business_requirement_deliverable_cost/models/business.py
```diff
@@ -185,6 +185,29 @@ class BusinessRequirement(models.Model):
...
+        readonly=True,
+        compute='_compute_get_ci_currency'
+    )
+
+    @api.multi
+    def _compute_get_ci_currency(self):
+        for br in self:
+            br.ci_currency_id = self.env.user.company_id.currency_id.id
+
+    @api.depends('total_revenue')
```
martintorresvm, line 204: @YogeshMahera-SerpentCS should be depend on this `currency_id` field so if change will be computed.

business_requirement_deliverable_cost/models/business.py
```diff
@@ -200,21 +223,30 @@ def _compute_rl_total_cost(self):
...
             if br.deliverable_lines:
-                br.resource_task_total =\
+                resource_task_total =\
                     sum(br.mapped('deliverable_lines').mapped('resource_ids')
                         .filtered(lambda r: r.resource_type == 'task')
                         .mapped('price_total'))
+                resource_task_total_ci = self.env['res.currency']._compute(
+                    br.currency_id, self.env.user.company_id.currency_id,
+                    resource_task_total)
+                br.resource_task_total = resource_task_total_ci
```
martintorresvm, line 233: @YogeshMahera-SerpentCS Try not use intermediate variable `resource_task_total_ci` just apply the new calculation.

Multiple places I've see so please have a check

martintorresvm (changes requested):

elicoidal:
@YogeshMahera-SerpentCS this should be done in v10, no need to fix version 8

YogeshMahera-SerpentCS:
@elicoidal  Noted!

PR #258, opened 2017-09-29
[ADD] account-analytic

martintorresvm:
- Would be need if might need
  for instance the parenting projects
- account_analytic_parent module


martintorresvm:
@elicoidal 

cc @YogeshMahera-SerpentCS

elicoidal (approved):
FT

PR #266, opened 2017-11-14
[IMP] [10.0] BR001208-T24350: Add field "Terms and Conditions"

JoJoJoJoJoJoJo:
add field `terms_and_conditions` and amend the printout

### DIFF
--- a/business_requirement/models/business.py
+++ b/business_requirement/models/business.py
@@ -67,6 +67,12 @@ class BusinessRequirement(models.Model):
         states={'draft': [('readonly', False)],
                 'confirmed': [('readonly', False)]}
     )
+    terms_and_conditions = fields.Html(
+        'Terms and Conditions',
+        read_only=True,
+        states={'draft': [('readonly', False)],
+                'confirmed': [('readonly', False)]}
+    )
     category_ids = fields.Many2many(
         'business.requirement.category',
         string='Categories',
--- a/business_requirement/models/business.py
+++ b/business_requirement/models/business.py
@@ -69,7 +69,7 @@ class BusinessRequirement(models.Model):
     )
     terms_and_conditions = fields.Html(
         'Terms and Conditions',
-        read_only=True,
+        readonly=True,
         states={'draft': [('readonly', False)],
                 'confirmed': [('readonly', False)]}
     )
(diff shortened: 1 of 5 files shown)


business_requirement/models/business.py
```diff
@@ -67,6 +67,12 @@ def _get_default_company(self):
         states={'draft': [('readonly', False)],
                 'confirmed': [('readonly', False)]}
     )
+    terms_and_conditions = fields.Html(
+        'Terms and Conditions',
+        read_only=True,
```
elicoidal: why readonly?
Shouldnot we add copy=true as well?

business_requirement/views/business_view.xml
```diff
@@ -63,6 +63,9 @@
                             <group string="Test Case">
                                 <field name="test_case" nolabel="1" default_focus="0"/>
```
elicoidal: remove default_focus="0"

elicoidal (changes requested):
LGTM: some details

Good job!

JoJoJoJoJoJoJo:
@elicoidal 
changes done by your request.

business_requirement/models/business.py
JoJoJoJoJoJoJo: default copy=True

elicoidal (changes requested):
almost perfect!
On the functional tests:
* Only thing missing: the field should not be printed in the printout if it is empty.

JoJoJoJoJoJoJo:
> Only thing missing: the field should not be printed in the printout if it is empty.

@elicoidal 
Any details about the test condition? I tested locally and it works normally

update:
I've found the situation, it seems when you create a BR, `<p><br></p>` will be added in the field, it's invisible but not empty

elicoidal:
@JoJoJoJoJoJoJo I understand then it is LGTM!
Good job :)

elicoidal (approved):
LGTM

elicoidal:
cc @victormartinelicocorp

lonelysun (approved):

martintorresvm (changes requested):
bump module version, the rest LGTM

JoJoJoJoJoJoJo:
@victormartinelicocorp 
version bumped
cc @elicoidal @lonelysun

martintorresvm (approved):

PR #267, opened 2017-11-15
layout_br_base doesn't wrap long lines

sysadminmatmoz:
It's not the perfect fix, since the page width has to be hard-coded to make word-wrap work, but since the default format you use is already A4, this should suffice to at least make the report usable, until we make t-if paper-format or something.
Another issue with wrapping is trying to fit it in a table cell, so this version gives the story, scenario, gap and test case in a normal A4 layout - actually I think this way suits the report better.

### DIFF
--- a/business_requirement_deliverable_report/views/report_business_requirement.xml
+++ b/business_requirement_deliverable_report/views/report_business_requirement.xml
@@ -105,63 +105,24 @@
             </table>
 
             <t t-if="o.business_requirement">
-                <h3>Customer Story:</h3>
-                <table class="table table-bordered">
-                    <tbody>
-                        <tr>
-                            <td>
-                                <pre style="padding-left: 30px;">
-                                    <span t-field="o.business_requirement"/>
-                                </pre>
-                            </td>
-                        </tr>
-                    </tbody>
-                </table>
+                <h3 style="width: 210mm; word-wrap: break-word;">Customer Story:</h3>
+                <span t-field="o.business_requirement"
+                      style="padding-left: 30px; width: 210mm; word-wrap: break-word; height: 100%;"/>
             </t>
 
             <t t-if="o.scenario">
-                <h3>Scenario:</h3>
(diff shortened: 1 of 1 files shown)


business_requirement_deliverable_report/views/report_business_requirement.xml
```diff
@@ -105,63 +105,24 @@
...
-                        <tr>
-                            <td>
-                                <pre style="padding-left: 30px;">
-                                    <t t-raw="o.test_case"/>
-                                </pre>
-                            </td>
-                        </tr>
-                    </tbody>
-                </table>
+                <span t-field="o.test_case" style="width: 210mm; word-wrap: break-word; height: 100%;"/>
```
sysadminmatmoz: Ok, here I missed a line in the commit before the span:
h3 style="width: 210mm; word-wrap: break-word;">Test Case: /h3

elicoidal:
@sysadminmatmoz thanks for your contribution!
I have compared with current version and actually cannot see improvement, rather the opposite.
In the following 2 screenshots, left is your version, right is current version of printout:
screenshot from 2017-11-16 08-26-29

screenshot from 2017-11-16 08-26-16

Is there a way to bring both wolds together?

sysadminmatmoz:
Hmm. Interesting... Completely different result at my end. Long lines don't get wrapped with the current version. Could it be something with the wkhtmltopdf version or maybe web_ckeditor4? I'll look into it some more and make a separate module with my version just for my existing instances meanwhile since they aren't usable as they are right now.
I also intend to find a way to integrate my existing change management module with this set (i need to keep the compatibility keeping in mind all my existing DB instances).
Disregard this PR for now and we'll be in touch :smile: 
Take care.

elicoidal:
@sysadminmatmoz it is strange indeed: to be noted that both have been issued using the current OCA runbot.
For this time I will close your PR hoping that you have the possibility to contribute later on :smile:

PR #309, opened 2019-06-04
[MIG] business_requirement: Migration to 11.0

martintorresvm:
cc @Tecnativa @pedrobaeza

PR #311, opened 2019-06-06
[MIG] business_requirement_deliverable: Migration to 11.0

martintorresvm:
Work in progress. Depends in :

- [x] https://github.com/OCA/business-requirement/pull/309


martintorresvm:
cc @Tecnativa, @pedrobaeza @Tardo 

Pending for review and add reports

martintorresvm:
I forced small changes. Problem here is no way of test on Runbot until base module is merged

business_requirement_deliverable/models/business.py
```diff
@@ -0,0 +1,280 @@
...
+                   ('confirmed', 'Confirmed'),
+                   ('approved', 'Approved'),
+                   ('stakeholder_approval', 'Stakeholder Approval'),
+                   ('in_progress', 'In progress'),
+                   ('done', 'Done'),
+                   ('cancel', 'Cancel'),
+                   ('drop', 'Drop'),
+                   ],
+        store=True,
+    )
```
sergio-teruel, line 31: This field is related without readonly, are you dure?
I do not remember if in v11 all related fields are readonly=true by default

business_requirement_deliverable/models/business.py
```diff
@@ -0,0 +1,280 @@
...
+    name = fields.Text('Name', required=True)
+    user_case = fields.Html()
+    proposed_solution = fields.Html()
+    product_id = fields.Many2one(
+        comodel_name='product.product',
+        string='Product',
+        domain=[('sale_ok', '=', True)],
+        required=False
```
sergio-teruel, line 39: Remove this attribute, by default is false

business_requirement_deliverable/models/business.py
```diff
@@ -0,0 +1,280 @@
...
+    )
+    uom_id = fields.Many2one(
+        comodel_name='product.uom',
+        string='UoM',
+        required=True,
+        default=lambda self: self.env.ref('product.product_uom_unit')
+    )
```
sergio-teruel, line 46: If you use a external id in default you need add an init hook and create the field manually

business_requirement_deliverable/models/business.py
```diff
@@ -0,0 +1,280 @@
...
+        string='Business Requirement',
+        ondelete='cascade',
+        required=True
+    )
+    sale_price_unit = fields.Float(
+        string='Sales Price',
+        oldname='unit_price'
+    )
+    price_total = fields.Float(
+        compute='_compute_get_price_total',
```
sergio-teruel: As a convention, _compute_price_total

business_requirement_deliverable/models/business.py
```diff
@@ -0,0 +1,280 @@
...
+        store=True,
+        readonly=True
+    )
+    currency_id = fields.Many2one(
+        comodel_name='res.currency',
+        string='Currency',
+        readonly=True,
+        compute='_compute_get_currency',
+    )
+    business_requirement_partner_id = fields.Many2one(
```
sergio-teruel, line 81: Why not use only partner_id

business_requirement_deliverable/models/business.py
```diff
@@ -0,0 +1,280 @@
...
+        comodel_name='res.partner',
+        related='business_requirement_id.partner_id',
+        string='Stakeholder',
+        store=True
+    )
+    business_requirement_project_id = fields.Many2one(
```
sergio-teruel: Project_id??

business_requirement_deliverable/models/business.py
```diff
@@ -0,0 +1,279 @@
...
+from odoo.exceptions import UserError
+from odoo.addons import decimal_precision as dp
+
+
+class BusinessRequirementDeliverable(models.Model):
+    _name = "business.requirement.deliverable"
+    _inherit = ["mail.thread", "mail.activity.mixin"]
+    _description = "Business Requirement Deliverable"
+
+    sequence = fields.Integer('Sequence')
```
pedrobaeza, line 18: Missing `_order = "sequence,id"`

business_requirement_deliverable/models/business.py
```diff
@@ -0,0 +1,279 @@
...
+        selection=[('draft', 'Draft'),
+                   ('confirmed', 'Confirmed'),
+                   ('approved', 'Approved'),
+                   ('stakeholder_approval', 'Stakeholder Approval'),
+                   ('in_progress', 'In progress'),
+                   ('done', 'Done'),
+                   ('cancel', 'Cancel'),
+                   ('drop', 'Drop'),
+                   ],
+        store=True,
```
pedrobaeza, line 29: Why is this needed? It requires `readonly=True`

martintorresvm: Is to control others fields on views to be editable but you are right need to be readonly true

martintorresvm:
Changes done

business_requirement_deliverable/models/business.py
```diff
@@ -0,0 +1,398 @@
...
+    )
+    business_requirement_partner_id = fields.Many2one(
+        comodel_name='res.partner',
+        related='business_requirement_id.partner_id',
+        string='Stakeholder',
+        store=True
+    )
+    state = fields.Selection(related='business_requirement_id.state',
+                             string='State', store=True, readonly=True)
+    portal_published = fields.Boolean('In Portal', default=True)
```
sergio-teruel, line 89: Why not use website_published and inherit from website.published.mixin

business_requirement_deliverable/models/business.py
pedrobaeza: In v11 default is `readonly=False`, so we have to explicitly put `readonly=True`

business_requirement_deliverable/models/business.py
pedrobaeza: Not this time, @sergio-teruel. That only happens when the external XML-ID is loaded in the same module. As this reference belongs to another module, there's no problem.

business_requirement_deliverable/models/business.py
pedrobaeza: Because with that, it will depend on `website` module, which is a heavy dependency.

business_requirement_deliverable/models/business.py
pedrobaeza: I suppose that this is because it's a related field.

business_requirement_deliverable/views/business_view.xml
```diff
@@ -0,0 +1,224 @@
...
+        <field name="arch" type="xml">
+            <form string="Business Requirement Task Type">
+                <sheet>
+                    <div class="oe_button_box" name="button_box" groups="base.group_user">
+                         <button class="oe_inline oe_stat_button" name="portal_publish_button" type="object" icon="fa-globe">
+                             <label for="portal_published" />
+                             <field name="portal_published" widget="toggle_button"/>
+                         </button>
+                    </div>
+                    <group>
```
chienandalu, line 121: The title is going to better hanging from the `<sheet>`

martintorresvm:
I pushed force dependency, monetary on Tree view, name search and group by section.

martintorresvm:
Please review @sergio-teruel @pedrobaeza

business_requirement_deliverable/views/business_view.xml
martintorresvm: Not sure how to do that, could you add suggested change?

Tardo:
Added dependencies in portal page

business_requirement_deliverable/views/business_view.xml
chienandalu:
```suggestion
```

business_requirement_deliverable/views/business_view.xml
```diff
@@ -0,0 +1,228 @@
...
+                        <div class="oe_title">
+                            <h1>
+                                <field name="name" attrs="{'readonly':[('state','not in',['draft', 'confirmed'])]}"/>
+                            </h1>
+                        </div>
+                    </group>
```
chienandalu, line 127:
```suggestion
```

business_requirement_deliverable/views/business_view.xml
```diff
@@ -0,0 +1,228 @@
...
+                <sheet>
+                    <div class="oe_button_box" name="button_box" groups="base.group_user">
+                         <button class="oe_inline oe_stat_button" name="portal_publish_button" type="object" icon="fa-globe">
+                             <label for="portal_published" />
+                             <field name="portal_published" widget="toggle_button"/>
+                         </button>
+                    </div>
+                    <group>
+                        <div class="oe_title">
+                            <h1>
```
chienandalu, line 123:
```suggestion
                            <h1 class="o_row">
```

business_requirement_deliverable/views/business_view.xml
```diff
@@ -0,0 +1,228 @@
...
+                                <field name="name" attrs="{'readonly':[('state','not in',['draft', 'confirmed'])]}"/>
```
chienandalu, line 124:
```suggestion
                                <field name="name" attrs="{'readonly':[('state','not in',['draft', 'confirmed'])]}" placeholder="Deliverable title..."/>
```

business_requirement_deliverable/views/business_view.xml
chienandalu: (it's not possible to add multiline suggestions, so you'll have to fix indent though)

rafaelbn:
Depends in:

- [ ] https://github.com/OCA/business-requirement/pull/309

martintorresvm:
Rebased

business_requirement_deliverable/views/business_view.xml
martintorresvm: get it it's about the indent

rafaelbn (approved):
Using it already in production without issues :+1: 

Maybe some improvements are needed in the future

PR #315, opened 2019-06-26
[MIG] business_requirement: Migration to v12

martintorresvm:
cc @Tecnativa

PR #317, opened 2019-07-02
[FIX] business_requirement: Bug when modify state confirmed-approved

martintorresvm:
cc @Tecnativa

#315

### DIFF
--- a/business_requirement/models/business.py
+++ b/business_requirement/models/business.py
@@ -199,11 +199,11 @@ class BusinessRequirement(models.Model):
         if vals.get('state'):
             br_xml_id = self.env.ref(
                 'business_requirement.group_business_requirement_manager')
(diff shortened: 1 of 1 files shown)


business_requirement/models/business.py
```diff
@@ -199,7 +199,7 @@ def write(self, vals):
         if vals.get('state'):
             br_xml_id = self.env.ref(
                 'business_requirement.group_business_requirement_manager')
-            user = self.env['res.users']
+            user = self.env.user.id
```
pedrobaeza: If you are putting an ID, call the variable `user_id`, but better to always store the recordset in variables and only populate ID when needed through `user.id`

martintorresvm:
Changes done @pedrobaeza

pedrobaeza (approved):

PR #318, opened 2019-07-15
[11.0] business_requirement_*: Sections improvement

pedrobaeza:
* Demo data
* Selectable on quotation wizard

@Tecnativa

### DIFF
--- a/business_requirement_sale/wizard/business_requirement_create_sale.py
+++ b/business_requirement_sale/wizard/business_requirement_create_sale.py
@@ -2,7 +2,7 @@
 # Copyright 2019 Tecnativa - Pedro M. Baeza
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
-from odoo import api, models, fields, _
+from odoo import _, api, exceptions, fields, models
 
 
 class BusinessRequirementCreateSale(models.TransientModel):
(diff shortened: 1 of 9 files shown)


business_requirement_sale/wizard/business_requirement_create_sale.py
```diff
@@ -36,17 +52,46 @@ def default_get(self, fields):
...
+                new_deliverable_ids += deliverables.ids
+        self.deliverable_ids = [(6, 0, new_deliverable_ids)]
+
+    @api.onchange('undefined_section')
+    def _onchange_undefined_section(self):
+        self.ensure_one()
+        deliverables = self.business_requirement_id.deliverable_lines.filtered(
+            lambda x: not x.section_id
+        )
+        command = 4 if self.undefined_section else 3
```
martintorresvm, line 92: I would add a comment, explains that deliverables will be added if undefined_section is True or removed if False
```suggestion
        # If undefined_section deliberables with no section will be added
        # otherwise need to be removed
        command = 4 if self.undefined_section else 3
```

pedrobaeza: But this is explaining how m2m commands work, which is not the scope of this code...

martintorresvm: I just feel better explaining what it does the function if the function name is not declarative. If could be as a docstring would be fine. Nevertheless is a nice to have

rafaelbn (approved):

martintorresvm (approved):
LGTM and tested on Runbot

Issue #321, opened 2019-08-07
[11.0][12.0] Access rights integration with website module


martintorresvm:
Steps to reproduce with any user which has no business requirement and portal groups
Go to menu: website>website>My account

Tested on Runbot 11.0:
```
403: Forbidden
The page you were looking for could not be authorized.
```

On 12.0:
Not yet as no module depends on website on the current migration status but as soon installs website app manually it happens.

martintorresvm:
cc @Tecnativa

pedrobaeza:
Please add the PR with the fix

PR #322, opened 2019-08-07
[FIX] business_requirement: Portal access on internal users

martintorresvm:
#321 cc @Tecnativa


pedrobaeza (changes requested):
The patch is invalid: you can't put that permission to all the users!!

As @Tardo did this part, better than he check this. Please superseed this.

martintorresvm:
Totally agree but IMHO is an Odoo feature. As soon you install `website` you will have menu Website for any internal user then if you install other modules with portal business objects like `website_sale` the access is granted at installation.
I made a functional test if remove sale groups on an internal demo user then the same kind of issue will happen when try access /my 

Any help is welcome if @Tardo could?

PR #325, opened 2019-08-20
[MIG] business_requirement_sale: Migration to v12

martintorresvm:


business_requirement_sale/static/description/index.html
```diff
@@ -3,7 +3,7 @@
 <html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
 <head>
 <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
-<meta name="generator" content="Docutils 0.15.1: http://docutils.sourceforge.net/" />
+<meta name="generator" content="Docutils 0.14: http://docutils.sourceforge.net/" />
```
pedrobaeza: Please regenerate this with newer version

business_requirement_sale/tests/test_business_requirement_sale.py
```diff
@@ -118,9 +118,9 @@ def test_flow_one_line_non_totaled(self):
         action = self.wizard.button_create()
         self.assertTrue(action['res_id'])
         order = self.env['sale.order'].browse(action['res_id'])
-        self.assertEqual(len(order.order_line), 1)
-        self.assertAlmostEqual(order.order_line.product_uom_qty, 3)
-        self.assertAlmostEqual(order.order_line.price_unit, 80)
+        self.assertEqual(len(order.order_line), 2)
+        self.assertAlmostEqual(order.order_line[1].product_uom_qty, 3)
```
pedrobaeza: Always filter by specific field, not by index, as order can be different

business_requirement_sale/views/sale_order_views.xml
```diff
@@ -22,7 +22,7 @@
         <field name="model">sale.order</field>
         <field name="inherit_id" ref="sale.view_sales_order_filter"/>
         <field name="arch" type="xml">
-            <field name="product_id" position="after">
+            <field name="order_line" position="after">
```
pedrobaeza: Why putting this data after the one2many instead of other more suitable place?

business_requirement_sale/wizard/business_requirement_create_sale.py
```diff
@@ -92,35 +92,30 @@ def _onchange_undefined_section(self):
...
         self.deliverable_ids = [(command, x) for x in deliverables.ids]
 
-    def _prepare_sale_layout_category_vals(self, deliverable, line_vals):
-        return {
-            'name': deliverable.section_id.name,
-        }
-
-    def _prepare_sale_order_line_vals(self, deliverable, order_vals):
+    def _prepare_sale_order_line_vals(
+            self, deliverable, order_vals, is_section=False):
```
pedrobaeza: Don't use the same method for sections than for sales order lines. Put a different one. You must make sure also through a sequence than the section is before the lines it contains, and also that all lines with same section are together.

business_requirement_sale/wizard/business_requirement_create_sale.py
```diff
@@ -136,7 +131,16 @@ def _create_sale_order(self):
         self.ensure_one()
         vals = self._prepare_sale_order_vals()
         for deliverable in self.deliverable_ids:
```
pedrobaeza: You must group deliverables by section first, and then create the section and append it, and then add its lines, all with a counter for the sequence.

pedrobaeza (changes requested):

pedrobaeza:
A migration script is also needed, but I will take care once comments are attended.

business_requirement_sale/wizard/business_requirement_create_sale.py
martintorresvm: Is it totally required to have a counter and the right sequence? I tested and the lines are in order as far I could check

martintorresvm:
I pushed new changes @pedrobaeza there is one comment which I am not sure I did as you required

business_requirement_sale/wizard/business_requirement_create_sale.py
pedrobaeza: What is the `_order` of the deliverable model? If it includes `section_id`, then they will be ordered when accessing `self.deliverable_ids`. If not, you need to sort them. But the sequence is needed for ensuring that no order criteria alters your order.

business_requirement_sale/wizard/business_requirement_create_sale.py
```diff
@@ -135,10 +132,31 @@ def _prepare_sale_order_vals(self):
...
+                    section, vals)
+                order.write({
+                    'order_line': [(0, 0, line_vals)],
+                })
+                for deliverable in deliverable_sections.filtered(
+                    lambda s: s.section_id.id == section.id
+                ):
+                    line_vals = self._prepare_sale_order_line_vals(
+                        deliverable, vals)
+                    order.write({
```
pedrobaeza: Why are you doing several write on the order? This reduces the performance a lot. And you haven't still set the sequence field I mentioned. You must prepare everything in the same loop, ordering deliverables by section and the rest of criteria (not grouping), and then when you detect a change in the section, calling the prepare section method for adding it before, with the corresponding sequence.

martintorresvm:
Changes done @pedrobaeza

business_requirement_sale/wizard/business_requirement_create_sale.py
```diff
@@ -92,16 +92,23 @@ def _onchange_undefined_section(self):
...
+            'display_type': 'line_section',
         }
+        return vals
 
-    def _prepare_sale_order_line_vals(self, deliverable, order_vals):
+    def _prepare_sale_order_line_vals(
+            self, deliverable, sequence, order_vals):
         vals = {
             'name': deliverable.name,
-            'product_id': deliverable.product_id.id,
```
pedrobaeza: Why changing this order in the dict? If you want to put them by alphabetical order, then you must sort `sequence` as well.

business_requirement_sale/wizard/business_requirement_create_sale.py
```diff
@@ -134,10 +133,29 @@ def _prepare_sale_order_vals(self):
 
     def _create_sale_order(self):
         self.ensure_one()
+        counter = 0
         vals = self._prepare_sale_order_vals()
-        for deliverable in self.deliverable_ids:
-            line_vals = self._prepare_sale_order_line_vals(deliverable, vals)
-            vals['order_line'].append((0, 0, line_vals))
+        deliverable_sections = self.deliverable_ids.filtered('section_id')
```
pedrobaeza: Don't do this filter. Just loop on deliverables. Some pseudo-code:
```
ant_section = False
sequence = 0
for deliverable in self.deliverable_ids:
    if deliverable.section_id != ant_section:
        sequence += 1
        line_vals = self._prepare_sale_order_line_section_vals(...)
        ant_section = deliverable.section_id
    sequence += 1
    line_vals = self._prepare_sale_order_line_vals(...)
```

pedrobaeza (changes requested):

martintorresvm:
Changes done @pedrobaeza

business_requirement_deliverable/models/business.py
```diff
@@ -11,7 +11,7 @@ class BusinessRequirementDeliverable(models.Model):
     _name = "business.requirement.deliverable"
     _inherit = ["mail.thread", "mail.activity.mixin", "portal.mixin"]
     _description = "Business Requirement Deliverable"
-    _order = "business_requirement_id, section_id, sequence, id"
+    _order = "business_requirement_id, section_id desc, sequence, id"
```
pedrobaeza: Why changing this?

business_requirement_sale/wizard/business_requirement_create_sale.py
```diff
@@ -135,8 +134,18 @@ def _prepare_sale_order_vals(self):
     def _create_sale_order(self):
         self.ensure_one()
         vals = self._prepare_sale_order_vals()
+        previous_section = False
+        sequence = 0
         for deliverable in self.deliverable_ids:
-            line_vals = self._prepare_sale_order_line_vals(deliverable, vals)
+            if deliverable.section_id and deliverable.section_id != previous_section:
```
pedrobaeza: Don't do here `if deliverable.section_id`, but inside:
```python
    if deliverable.section_id != previous_section:
        if deliverable.section_id:
            sequence += 1
            line_vals = self._prepare_sale_order_line_section_vals(
                deliverable.section_id, sequence)
            vals['order_line'].append((0, 0, line_vals))
        previous_section = deliverable.section_id
```

business_requirement_deliverable/models/business.py
pedrobaeza: Check my other comment that solves this.

martintorresvm: As far I understand it is needed to create first the deliverable lines with undefined section.
Please check this screenshot:
screenshot-127 0 0 1_12069-2019 08 22-19_48_03
But is inside the section Optional

Was the main reason I did with filtered method, but with the order you show was possible done on the same loop

pedrobaeza: OK, I see, I haven't though on this. We have 3 possible solutions:

- Create a section line "[No section]" (translated).
- Call `br.deliverable_ids.sorted(...` with that order for having first the ones with no section. But I don't like it because the rest of the sections are not respected. I'm also thinking that we have a sequence in sections, isn't it? (or if not, we should), for sorting how do we want to put sections. This is not respected when we put `section_id` in the order, but there's no way to do it in the `_order` attribute. We must do it through the sorted function. We can specify the sorting method for putting first the no section and then sort by section sequence.
- Split in 2 the loop, but we duplicate code. One variant is to pre-sort manually on a populated recordset our deliverables and then do the loop, but previous sorted custom method seems cleaner.

martintorresvm: sorted method works fine by default is getting the undefined sections then first section in order though we are done here except for the migration script

martintorresvm:
Changes done @pedrobaeza, see that I follow the recommendation of use sorted method.
And the section order now is by sequence, I had to added. Results seems correct to me.

martintorresvm:
Now test improved, check if order by section sequence is correct having undefined section first
cc @pedrobaeza

pedrobaeza (approved):
I have improved tests + other fixes

PR #343, opened 2021-03-07
[FIX] business_requirement_deliverable: Count unpublished deliverables in portal

sergio-teruel:
cc @Tecnativa
ping @CarlosRoca13 @victoralmau

### DIFF
--- a/business_requirement_deliverable/models/business.py
+++ b/business_requirement_deliverable/models/business.py
@@ -168,7 +168,10 @@ class BusinessRequirement(models.Model):
     dl_total_revenue = fields.Float(
         string="DL Total Revenue", digits="Account", compute="_compute_dl_total_revenue"
     )
-    dl_count = fields.Integer("DL Count", compute="_compute_dl_count")
+    dl_count = fields.Integer(string="DL Count", compute="_compute_dl_count")
+    dl_count_portal_published = fields.Integer(
+        string="DL Count Portal pubished", compute="_compute_dl_count_portal_published"
+    )
     pricelist_id = fields.Many2one(
         comodel_name="product.pricelist",
         string="Pricelist",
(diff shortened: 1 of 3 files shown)


victoralmau (approved):

business_requirement_deliverable/models/business.py
```diff
@@ -194,9 +194,13 @@ def _compute_dl_total_revenue(self):
         for r in self:
             r.dl_total_revenue = sum(r.deliverable_lines.mapped("price_total"))
 
+    @api.depends_context("website_id")
     def _compute_dl_count(self):
         for r in self:
-            r.dl_count = len(r.deliverable_lines.ids)
+            if "website_id" in self.env.context:
```
pedrobaeza: I don't see why you need the website here, as this doesn't depend on website nor the field is depending on it.

CarlosRoca13 (approved):

marcelsavegnago (approved):

pedrobaeza (changes requested):
Please check my comment

sergio-teruel:
Why have I done this?
For one business requirement (BR) published in portal you can published all BR deliveries or any, so in BR portal view we van not show all deliveries counter, we only can display the published deliveries but this behaviour only is for portal website, in backed we want count all deliveries...

image

pedrobaeza:
Ok, I get the goal, but I'm afraid the implementation is not correct, as you can have portal without website, so you need to create a new field to be used in this part and not to rely on such context.

sergio-teruel:
Ok... I will re-check so...

victoralmau:
Finally I have added a new field that already filters the records published on the website.
If this change is correct, we will add it in v14 as well.
What do you think about it?

pedrobaeza (approved):

PR #345, opened 2021-10-14
[14.0][MIG] business_requirement: Migration to 14.0

victoralmau:
Migration to 14.0

Locked by:
- https://github.com/OCA/business-requirement/pull/346

Please @joao-p-marques and @pedrobaeza can you review it?

@Tecnativa TT30253


business_requirement/models/business.py
```diff
@@ -152,23 +147,26 @@ def _get_default_company(self):
     )
     portal_published = fields.Boolean("In Portal", default=False)
     user_id = fields.Many2one(
-        "res.users",
+        comodel_name="res.users",
         string="Owner",
         default=lambda self: self.env.user,
         required=True,
-        track_visibility="always",
+        tracking=True,
```
CarlosRoca13: If this is migrated from 13.0 this should be fixed on that version too. Because it is probably having no effect

CarlosRoca13 (approved):
LGTM :+1: I will add a PR to avoid the problem on v13

business_requirement/models/business.py
victoralmau: Thanks for creating the PR for v13, when it is merged I will add it here, reducing the difference of changes.

pedrobaeza (approved):

business_requirement/views/business_view.xml
```diff
@@ -159,7 +159,7 @@
                                 <div class="oe_kanban_bottom_right">
                                     <div class="pull-left" groups="base.group_user">
                                         <img
-                                            t-att-src="kanban_image('res.users', 'image_small', record.responsible_user_id.raw_value)"
+                                            t-att-src="kanban_image('res.users', 'image_128', record.responsible_user_id.raw_value)"
```
pedrobaeza, line 162: I'm now doubting. Shouldn't this be `image_64`?

victoralmau: The smallest size is `image_128` (`image_64` does not exist).

pedrobaeza: OK, but if the size is not the same as previous, then maybe a manual size parameter should be added.

victoralmau: The `width` and `height` parameters were previously defined.

https://github.com/OCA/business-requirement/blob/8263e15bfaf5c360ff75e08080e62242ce318f4b/business_requirement/views/business_view.xml#L161-L168

pedrobaeza: Ah, OK

PR #346, opened 2021-10-14
[13.0] [IMP] business_requirement: Change track_visibility by tracking

CarlosRoca13:
Since 13.0 the key track_visibility has no effect so should be changed on
the migration by tracking=True

ping @pedrobaeza @victoralmau 

@Tecnativa


pedrobaeza (approved):
I think it still works, but good to not use deprecated things

victoralmau (approved):

PR #350, opened 2021-11-11
[14.0][MIG] business_requirement_deliverable: Migration to 14.0

victoralmau:
Migration to 14.0

Please @CarlosRoca13  and @joao-p-marques can you review it?

@Tecnativa TT30255


business_requirement_deliverable/models/business.py
```diff
@@ -82,7 +82,7 @@ def _compute_currency_id(self):
             if currency_id:
                 brd.currency_id = currency_id
             else:
-                brd.currency_id = self.env.user.company_id.currency_id
+                brd.currency_id = self.env.company.currency_id
```
joao-p-marques: This might actually change behavior, right?

- `self.env.user.company_id` would be the default company associated with the user (and this field still exists in v14
- `self.env.company` would be the main company activated in the environment

Can this change in behavior be a problem?

business_requirement_deliverable/models/business.py
```diff
@@ -227,7 +227,7 @@ def _compute_currency_id(self):
             if br.partner_id and br.pricelist_id.currency_id:
                 br.currency_id = br.pricelist_id.currency_id.id
             else:
-                br.currency_id = self.env.user.company_id.currency_id.id
+                br.currency_id = self.env.company.currency_id.id
```
joao-p-marques: Same here as above

joao-p-marques (commented):
Code review

business_requirement_deliverable/models/business.py
victoralmau: In my opinion it is better to use `self.env.company`.
What do you think your @pedrobaeza ?

pedrobaeza: Isn't there another company available right there in the model? If not, the change seems good for me and in fact it should be changed in v13, not here (but no need to be backported).

business_requirement_deliverable/models/business.py
victoralmau: It is changed to use the company value from the record.

business_requirement_deliverable/models/business.py
victoralmau: I have changed it to use the value of `business_requirement_id.currency_id` from the record.

joao-p-marques: Ok, looks good :+1:

joao-p-marques (approved):

victoralmau:
Can you retry the bot process?

pedrobaeza:
I haven't launched any merge. I have marked the PR in the migration issue. It's the new command that is already deployed (you have worked on it as well).

Tardo (approved):
LGTM. Code review

PR #395, opened 2026-07-24
[17.0][MIG] business_requirement_crm: Migration to 17.0

alan196:
Migration of `business_requirement_crm` from 16.0 to 17.0.

Depends on #394 (`business_requirement` migration).

Standard migration performed with `oca-port` (full commit history preserved), plus the following API adaptations for 17.0:

- Converted the two `attrs=` usages in `crm_lead_views.xml` to direct `invisible` expressions.
- Fixed the `_read_group()` call in `_compute_business_requirement_count` to use the new 17.0 API (tuple results with recordset keys).
- `open_requirements()`: switched to `ir.actions.actions._for_xml_id()` and fixed the multi-record `domain` value, which was wrapped in a tuple and produced an invalid action domain.
- Tests: `SavepointCase` → `TransactionCase`.
- Removed a stray `%s` placeholder in the wizard chatter message that rendered the raw values dict inside the message text.

The existing test passes and pre-commit is clean.


szalatyzuzanna:
Hi @alan196, are you still working on this one? I'm preparing the migration of this module to 18.0 and wanted to check with you first, in case you also had plans for that version.

alan196:
Hello, @szalatyzuzanna we already migrated the module to 19.0, we won't be working in this PR

PR #397, opened 2026-09-10
[18.0][MIG] business_requirement: Migration to 18.0

szalatyzuzanna:
Migration of `business_requirement` from 16.0 to 18.0, starting from 16.0 since the module was never migrated to 17.0 (see #394, still open).

Supersedes #390 (closed without merge, stuck on tests).

Standard migration with `git format-patch --keep-subject` (full commit history preserved), plus the following adaptations for 17.0 and 18.0:

- Removed the 15 `states=` field parameters; the state-based readonly behaviour moved to the form view as `readonly` expressions, keeping field-by-field parity with 16.0. `readonly=True` kept on the model for the audit fields only.
- Replaced `name_get()` with a `_compute_display_name` override, and the `name_search()` override with `_rec_names_search = ["name", "description"]`.
- Dropped the `read_group()` override in favour of `group_expand=True` on `state`.
- `message_post()` adapted to the keyword-only signature.
- Controller: `check_access_rights()` → `has_access()`, `render_qweb_pdf()` → `_render_qweb_pdf()`.
- Views: `tree` → `list`, `<div class="oe_chatter">` → `<chatter/>`, kanban on `t-name="card"`, `kanban_state_selection` → `state_selection`, `t-esc` → `t-out`. The removed `toggle_button` widget on the publish stat button is replaced by two `o_stat_text` spans, dropping the double write it caused.
- Portal: `portal.portal_record_layout` is gone in 18.0, record page rebuilt as a card; home entry moved to `portal_service_category`. Bootstrap 4 leftovers in that template fixed along the way.
- Tour: `odoo.define` → `web_tour.tours` registry, file renamed to `.esm.js` for the repo eslint config.
- Tests: `SavepointCase` → `TransactionCase`, plus coverage for `display_name`, the expanded `read_group` and the publish button.

@BinhexTeam


jelenapoblet:
@pedrobaeza we would like to continue with this project, since requirements is something that one could need internally too, to manage request from the company to the product dev team, i.e.
Also we'll problably continue to migrating and improving the requirements to SO too.
Let me know if it makes sense from your point of view.

pedrobaeza:
Well, we discard its usage for not fitting our needs, but you can go ahead.

jelenapoblet (approved):
Standard migration, tested on runboat, LGTM

oliverg09:
Great job @szalatyzuzanna ! Small inconsistency here: the block is conditioned on `br.user_id`, `_contactName`/`_contactAvatar` are taken from `br.responsible_user_id`, and the contact widget below renders `br.user_id` email/phone. In 16.0 the displayed name came from `user_id` (the contact widget included "name"), so this changes what the portal shows when owner and responsible differ. 

Could you align all three on the same user?

szalatyzuzanna:
@oliverg09 good catch, thanks for the careful review.

Fixed: name, avatar and contact details now all come from `user_id`, which is what 16.0 actually displayed, since the contact widget rendered `user_id` with "name" included. So portal users see exactly the same data as before the migration. Squashed into the `[MIG]` commit, as it was a regression I introduced while rewriting the template.

oliverg09 (approved):
Code review. LGTM!

PR #399, opened 2026-09-14
[18.0][ADD] business_requirement_rice: prioritise with RICE

szalatyzuzanna:
New module to prioritise business requirements with the RICE scoring model, for teams that collect more requests than they can work on and need to order them objectively.

Four estimates are combined into a stored score (`Reach x Impact x Confidence / Effort`), so requirements can be sorted, grouped and filtered by it:

- **Reach**: how many users or events are affected over a given period.
- **Impact**: 0.25 / 0.5 / 1 / 2 / 3, from minimal to massive.
- **Confidence**: 50% / 80% / 100%.
- **Effort**: total work, usually in person-months.

Some design notes, in case they come up in review:

- The fields live in their own `RICE` notebook page rather than in the main form, so that another prioritisation model can be added later as a separate page without crowding the requirement.
- The existing `priority` field is left untouched, and no default order is imposed on the list view: the module only adds fields, a page, a column and a search filter, and does not change the behaviour of the module it extends.
- An effort of zero or less scores zero instead of raising, since the field is free input.
- Impact and confidence use the standard fixed scales. Making them configurable would need a model of its own, which seemed disproportionate for a first version.

@BinhexTeam


jelenapoblet (approved):
Looks good to me. The RICE scoring is nicely isolated in a dedicated module and the implementation is simple and clean. Thanks!

oliverg09:
Good work overall @szalatyzuzanna. The RICE scoring logic is clean and the test coverage for edge cases (empty selections, zero/negative effort, sorting) is solid. Two small robustness suggestions, non-blocking:

1. `rice_impact`/`rice_confidence` selection keys are cast with `float()` — consider a weight dict instead, so a future key change degrades to 0 instead of raising `ValueError`.
2. `rice_effort <= 0` is guarded and forced to score 0, but `rice_reach` isn't — a negative reach would produce a negative score. Worth the same guard.

Nice work either way, thanks for the contribution!

szalatyzuzanna:
Thanks @oliverg09 , both applied.

**1.** Weights now live in module-level dicts with semantic selection keys, so the stored value identifies the estimate rather than its arithmetic value; the lookup raises on an unknown key instead of defaulting to zero, and `test_every_option_has_a_weight` makes a missing weight fail the tests rather than runtime.

**2.** Negative reach is guarded alongside the existing effort check, with `test_negative_reach_is_not_scored`.

oliverg09 (approved):
LGTM!

jelenapoblet:
@OCA/pypi-support Please reserve the name for this addon.

PR #400, opened 2026-09-14
[IMP] business_requirement: refresh module, menu and portal icons

szalatyzuzanna:
Refreshes the three icons of the module, each of which is shown in a different context and so far reused the same file.

**Module icon** (`static/description/icon.png`) — replaced with the current OCA icon. The previous one predates that convention and stood out in the apps list.

**Menu icon** (`static/description/menu_icon.png`, new) — the top menu had no icon of its own and fell back to the module one, which is not designed for that size. This adds a 140x140 PNG with transparency, which is what Odoo expects for `web_icon`.

**Portal icon** (`static/img/requirements.svg`, new) — the portal entry used `static/description/icon.svg`, a module icon rather than a portal pictogram. The new file is an SVG declaring 64x64, the size Odoo's own portal entries use, so it stays sharp on high-density displays.

`static/description/icon.svg` is intentionally left in place: it is still referenced by `business_requirement_deliverable`, whose portal entry will be updated in a separate PR.

@BinhexTeam


jelenapoblet (approved):
Nice Icons. Thank you!

PR #401, opened 2026-09-16
[18.0][IMP] business_requirement_deliverable: refresh module and portal icons

szalatyzuzanna:
Refreshes the two icons of the module, following #400, which did the same for `business_requirement`.

**Module icon** (`static/description/icon.png`) — replaced with the current OCA icon. The previous one predates that convention and stood out in the apps list.

**Portal icon** (`static/img/deliverables.svg`, new) — the portal entry used `business_requirement/static/description/icon.svg`, a module icon belonging to another addon, so both portal entries showed the same picture. This gives the deliverables entry its own icon, an SVG declaring 64x64, the size Odoo's own portal entries use, so it stays
sharp on high-density displays.

As a side effect the template no longer points at a file from `business_requirement`, so the two modules' portal icons are now independent.


jelenapoblet (approved):
Nice job

PR #398, opened 2026-09-10
[18.0][MIG] business_requirement_deliverable: Migration to 18.0

szalatyzuzanna:
Migration of `business_requirement_deliverable` from 14.0 to 18.0, starting from 14.0 since the module was never migrated to 15.0, 16.0 or 17.0.

Builds on top of `business_requirement` (#397, already merged).

Standard migration with `git format-patch --keep-subject` (full commit history preserved), plus the following adaptations for 15.0 → 18.0:

- Pricelist API: the `with_context(pricelist=...)` + `product.price` pattern is gone, replaced by `pricelist._get_product_price()`. Both `product_id_change()` and `product_uom_change()` had the same duplicated block, now factored into `_get_sale_price_unit()`.
- Removed the `states=` parameters and the `readonly=True` on computed fields; the state-based readonly behaviour moved to the form view as `readonly` expressions.
- `name_get()` → `_compute_display_name` override, `check_access_rights()` → `has_access()`.
- `copy()` / `copy_data()` adapted to the multi-record signature, keeping the follower and deliverable mapping per record.
- `message_subscribe()` / `message_unsubscribe()`: `channel_ids` dropped, `message_channel_ids` no longer exists. The subtype check is now a single `any()` instead of a loop that mutated the list it was iterating over.
- Views: `tree` → `list`, `<chatter/>`, kanban on `t-name="card"`, `view_type` / `tree_view_ref` → `list_view_ref`, `t-esc` → `t-out`. The removed `toggle_button` widget is replaced by `boolean_toggle` in the kanban and by two `o_stat_text` spans on the stat button.
- Portal: `portal.portal_record_layout` is gone in 18.0, record page rebuilt as a card; home entry moved to `portal_service_category`; backend URL to `/odoo/...`. Bootstrap 4 leftovers fixed in the portal templates and the report.
- `assets.xml` templates replaced by the manifest `assets` block; tour rewritten on the `web_tour.tours` registry with `run: "click"` and `expectUnloadPage`, renamed to `.esm.js` for the repo eslint config.
- `product.group_sale_pricelist` → `product.group_product_pricelist`.
- Demo data: the `product.template` + `product.product` pairs collapsed into single `product.product` records, they hit `product_product_combination_unique` otherwise.

@BinhexTeam

PR #402, opened 2026-09-20
[18.0][MIG] business_requirement_crm: Migration to 18.0

szalatyzuzanna:
Migration of `business_requirement_crm` from 16.0 to 18.0, starting from 16.0 since the module was never migrated to 17.0.

There is an open draft for the 17.0 migration (#395), untouched since July. This PR targets 18.0 directly, the same way `business_requirement` (#397) and `business_requirement_deliverable` (#398) were migrated.

Standard migration with `git format-patch --keep-subject` (full commit history preserved), plus the following adaptations for 17.0 → 18.0:

- `read_group()` → `_read_group()` with `aggregates=["__count"]`; the result is now an iterable of tuples, so the count dict is built from `(lead, count)` pairs instead of `x["lead_id_count"]`.
- `self.env.ref(...).read()[0]` → `_for_xml_id()` to read the action.
- `_()` imported from `odoo` is deprecated, replaced by `self.env._()` with keyword arguments instead of `%` formatting.
- `message_post()` escapes plain strings since 17.0, so the chatter links are built as `Markup` and passed as an argument. The translatable terms no longer carry any HTML, and the record names are escaped on interpolation.
- `default_get(fields)` → `default_get(fields_list)`.
- Views: the removed `attrs` replaced by `invisible` expressions.

One pre-existing bug fixed along the way: `open_requirements()` built the multi-record domain as `([("lead_id", "in", self.ids)],)`, a tuple wrapping the domain instead of the domain itself.

A second commit adds tests for `open_requirements()`, which had no coverage at all, and for the chatter links. Tests go from 1 to 4 and the module reaches 100% coverage on every Python file.

@BinhexTeam


jelenapoblet (approved):
Tested on runboat. Works as spected.

business_requirement_crm/wizard/crm_lead_convert_to_requirement.py
```diff
@@ -0,0 +1,77 @@
...
+        readonly=True,
+        domain=[("type", "=", "opportunity")],
+    )
+    description = fields.Char(required=True)
+    customer_history = fields.Html()
+
+    @api.model
+    def default_get(self, fields_list):
+        result = super().default_get(fields_list)
+        lead = self.env["crm.lead"].browse([self.env.context.get("active_id")])
```
popadron: Maybe worth guarding default_get from a missing active_id? Currently browse([None]) would still pass if lead:. Non-blocking, as this was already there before the migration.

business_requirement_crm/readme/USAGE.md
```diff
@@ -0,0 +1,4 @@
+To use this module, you need to:
+
+1.  Go to CRM Pipeline Lead menu
+2.  Click on Convert to Requirement button
```
popadron: In readme/USAGE.md the step says "Convert to Requirement", but the button in views/crm_lead_views.xml is labelled "Create Business Requirement". Might be worth aligning the two so the docs match the UI.

business_requirement_crm/wizard/crm_lead_convert_to_requirement_views.xml
```diff
@@ -0,0 +1,32 @@
...
+                    <field name="customer_history" />
+                </group>
+                <footer>
+                    <button
+                        type="object"
+                        name="action_lead_to_business_requirement"
+                        string="Create Requirement"
+                        class="btn-primary"
+                    />
+                    <button string="Cancel" class="btn-default" special="cancel" />
```
popadron: Since Odoo 18 uses Bootstrap 5, the Cancel button in the wizard view could use `btn-secondary` instead of `btn-default`.

popadron (approved):
LGTM

szalatyzuzanna:
@popadron Thanks, all three applied. They come unchanged from 16.0. Note the `default_get` one wasn't actually non-blocking: `browse([None])` returns a truthy recordset, so the `if lead:` guard never caught it.

PR #403, opened 2026-09-20
[18.0][MIG] business_requirement_sale: Migration to 18.0

szalatyzuzanna:
Migration of `business_requirement_sale` from 14.0 to 18.0, starting from 14.0 since the module was never migrated to 15.0, 16.0 or 17.0.

Builds on top of `business_requirement_deliverable` (#398, already merged).

Standard migration with `git format-patch --keep-subject` (full commit history preserved), plus the following adaptations for 15.0 → 18.0:

- `read_group()` → `_read_group()` with `aggregates=["__count"]`; the result is now an iterable of tuples, so the count dict is built from `(record, count)` pairs.
- `self.env.ref(...).read()[0]` → `_for_xml_id()` to read the action.
- `_()` imported from `odoo` is deprecated, replaced by `self.env._()`.
- `message_post()` escapes plain strings since 17.0, so the chatter links are built as `Markup` and passed as an argument. The translatable terms no longer carry any HTML, and the record names are escaped on interpolation.
- `default_get(fields)` → `default_get(fields_list)`.
- Views: the removed `attrs` replaced by `invisible` expressions, and  `groups_id` on the button replaced by the `groups` attribute. The sale order field moved after `team_id`, since `analytic_account_id` is no longer in that position on the 18.0 form.

Three pre-existing issues fixed along the way:

- There were two `SaleOrderLine` classes inheriting the same model, one in `models/sale_order.py` and one in `models/sale_order_line.py`, which `consider-merging-classes-inherited` rejects. Merged into  `sale_order.py` and the duplicate file removed.
- `open_orders()` built the multi-record domain as `([("business_requirement_id", "in", self.ids)],)`, a tuple wrapping the domain instead of the domain itself.
- The wizard's `business_requirement_id` carried `domain="[('business_requirement_id', '=', business_requirement_id)]"`, a field that does not exist on `business.requirement`. It was inert because the field is `invisible="1"` and always filled from `default_get`, but it would raise as soon as the field were shown. It looks like a copy of the domain on `deliverable_ids`, where the field does exist. Removed.

@BinhexTeam


business_requirement_sale/wizard/business_requirement_create_sale.py
```diff
@@ -0,0 +1,190 @@
...
+                    raise exceptions.UserError(
+                        self.env._(
+                            "No deliverables found for this business requirement."
+                        )
+                    )
+            if "has_undefined_section" in fields_list:
+                res["has_undefined_section"] = any(
+                    not x.section_id for x in br.deliverable_lines
+                )
+            if "section_ids" in fields_list:
```
popadron: The check is `section_ids` in fields_list, but the value being set is `applicable_section_ids`. It works because the form has both fields, is this intentional?. I think checking `applicable_section_ids` would make the intent clearer

business_requirement_sale/views/sale_order_views.xml
popadron, line 1: Just a small note: the team_id change seems to be in the search view (`view_sales_order_filter`), not the form as the description says. Is that right?

business_requirement_sale/tests/test_business_requirement_sale.py
```diff
@@ -0,0 +1,182 @@
...
+        order = self.env["sale.order"].browse(action["res_id"])
+        self.assertEqual(order.partner_id, self.partner)
+        self.assertEqual(len(order.order_line), 5)
+        self.assertEqual(self.business_requirement.sale_order_count, 1)
+        self.assertEqual(order.business_requirement_id, self.business_requirement)
+        action = self.business_requirement.open_orders()
+        self.assertTrue(action["context"]["search_default_business_requirement_id"])
+        br2 = self.business_requirement.copy()
+        action = (self.business_requirement + br2).open_orders()
+        self.assertTrue(action["domain"])
```
popadron: This assert might not catch a regression of the domain fix, since the old tuple-wrapped value is also truthy.
```suggestion
        self.assertEqual(
            action["domain"],
            [("business_requirement_id", "in", (self.business_requirement + br2).ids)],
        )
```

business_requirement_sale/tests/test_business_requirement_sale.py
popadron, line 1: Could you add a test for the chatter messages, like test_chatter_messages_keep_their_link? Since message_post escapes plain strings, a test checking that `<a` is in the body (and `&lt;a` is not) would make sure the Markup links aren't escaped in either the quotation or the requirement

popadron (approved):
LGTM

szalatyzuzanna:
@popadron Thanks, all four applied. Good catch on the domain assert — you're right that the tuple-wrapped value was also truthy, so the test wasn't guarding anything. Added the chatter test too, and verified it fails if the `Markup` is dropped. On the `team_id` note: you're right, it's the search view, not the form — I've fixed the description.

jelenapoblet (approved):
Functional review ok

